Issue #5, opened 2024-06-04
Web socket messages processed out of order


kfigiela:
Hey! I've been playing around with hyperbole for last couple of days. In my use-case, I have two hyper views polling similar to the example in https://github.com/seanhess/hyperbole/blob/main/example/Example/LazyLoading.hs. 

In client code in https://github.com/seanhess/hyperbole/blob/748444e8bfe68273fbb922f19a1674fc2b4a5176/client/src/sockets.ts#L78
the request is sent and it awaits the response, which is the first incoming message on the websocket. However, in case of multiple parallel requests, responses may be returned out-of-order. This causes one hyper view to be replaced with the contents of the other. 

Probably, there should be some request id that is returned by the response, so appropriate event handler is called on JS side.

Alternatively, the response could indicate what hyper view needs to be updated making this fully asynchronous. That'd open a possibility for server initiated push.

seanhess:
Hey, thanks for submitting this! This has been in the back of my mind for a
while.

The requests are targeted to a particular hyperview, so a simple solution
for this would be to have the response return the view id, and look up the
correct view in the response: make the request and the response
independent.

But that wouldn’t prevent errors if the client triggers another update
within the same view while a request is in flight: like if someone
double-clicks a submit button, or clicks a button while polling. These
could be prevented using onRequest, but it needs better default behavior.

For that we would need a request id. What’s the correct logic for out of
order responses? Just apply them in order? Only render the latest?

What are other ways that requests could overlap and mess things up?

kfigiela:
I did some experimentation and opened a [draft PR](https://github.com/seanhess/hyperbole/pull/6) to help with the discussion. By adding a correlation id, I was able to resolve the issue of updating incorrect hyperview. 

Out of order arrival of responses for the same hyper view was not an issue at the moment. Websocket connection was processed single-threaded, so the messages were processed sequentially on backend. This did not apply to HTTP fallback though, which would be processed multi-threaded and could be an issue. 

I did small refactor of how web sockets are handled in the backend and that added concurrency there which should improve UI responsiveness in case there are long responses. This still needs more work (e.g. to kill asyncs when connection is gone, etc.), but overall works. Since out-of-order responses are a thing now, I added a check on response processing to prevent rendering response if the target has been already updated by later request.

[...]

PR #6, opened 2024-06-06
Fix race conditions

kfigiela:
Attempts to fix various issues related to https://github.com/seanhess/hyperbole/issues/5. This is still a draft and requires more work.


src/Web/Hyperbole/Application.hs
```diff
@@ -47,24 +48,25 @@ import Web.View (View, renderLazyByteString, renderUrl)
...
 
 
-socketApp :: (IOE :> es) => Eff (Hyperbole : Server : es) Response -> PendingConnection -> Eff es ()
+socketApp :: (IOE :> es, Concurrent :> es) => Eff (Hyperbole : Server : es) Response -> PendingConnection -> Eff es ()
 socketApp actions pend = do
   conn <- liftIO $ WS.acceptRequest pend
   forever $ do
-    runServerSockets conn $ runHyperbole actions
+    msg <- liftIO $ WS.receiveData conn
+    void $ forkIO $ void $ runServerSockets conn msg $ runHyperbole actions
```
seanhess, line 64: I think we might be able to shoot ourselves in the foot with `forkIO` here. My read of the docs says it won't propogate exceptions, among other things. Let's read up on https://hackage.haskell.org/package/effectful-2.3.1.0/docs/Effectful-Concurrent-Async.html

src/Web/Hyperbole/HyperView.hs
```diff
@@ -32,6 +32,10 @@ instance HyperView Message where
 -}
 class (Param id, Param (Action id)) => HyperView id where
   type Action id :: Type
+  hyperViewMod :: id -> Mod
```
seanhess, line 35: Can you tell me how this came up? Could it have been solved by nesting a `col` inside of a `hyper`?

seanhess:
Closing in favor of #7, but this was super-helpful. I'll keep stealing ideas and code from this branch. Thanks

src/Web/Hyperbole/Application.hs
kfigiela: Absolutely, I added `forkIO` as a quick fix to see how this will affect and had this on my todo list. Async (effectful, or outside of effects machinery) sounds like a good solution. The question is how to react to exceptions, i.e. if this should kill the socket, etc.

PR #7, opened 2024-06-09
Concurrent: Out of order socket responses

seanhess:
Non-draft implementation of #6 

* Minimal Reproduce Example
* Server responds with view-ids
* Promises ignore responses not intended for them

Fixes the main issue described in #5

### DIFF
--- a/client/src/action.ts
+++ b/client/src/action.ts
@@ -1,4 +1,4 @@
-export function actionUrl(id:string, action:string):URL {
+export function actionUrl(id:ViewId, action:string):URL {
   let url = new URL(window.location.href)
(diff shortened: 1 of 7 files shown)


kfigiela:
Looks good. As noted in [the comment](https://github.com/seanhess/hyperbole/issues/5#issuecomment-2152113635) there are three outstanding issues to be addressed separately:

* web socket parallelism,
* possible out of order responses,
* race conditions related to ongoing requests/onLoad timers.

seanhess:
This should resolve all issues until we add web socket parallelism. We can fix anything that arises after adding parallelism in that issue/PR

PR #8, opened 2024-06-10
Hyperbole to allow arbitrary mods on hyper view

kfigiela:
Specifically, allow any CSS class to be set there and do not force col.

### DIFF
--- a/src/Web/Hyperbole/HyperView.hs
+++ b/src/Web/Hyperbole/HyperView.hs
@@ -32,6 +32,10 @@ instance HyperView Message where
 -}
 class (Param id, Param (Action id)) => HyperView id where
   type Action id :: Type
+  hyperViewMod :: id -> Mod
+  default hyperViewMod :: id -> Mod
+  hyperViewMod _ = id
+
 
 
 {- | Embed HyperViews into the page, or nest them into other views
@@ -65,7 +69,7 @@ otherView = do
 -}
 hyper :: forall id ctx. (HyperView id) => id -> View id () -> View ctx ()
 hyper vid vw = do
-  el (att "id" (toParam vid) . flexCol) $
+  el (att "id" (toParam vid) . hyperViewMod vid) $
     addContext vid vw
 
 


seanhess:
Thanks for opening a focused PR for this. As mentioned in the other issue, I have a question about this:

HTML and CSS are really great at using nesting for composability. I haven't yet found any use-case that couldn't be solved by nesting another container inside the `hyper`. 

```haskell
hyper MyView $ do
  -- use styles and layout here
  col (pad 20 . gap 20 . color Green) $ do
    el_ "Hello"
```

Would your issue be solved using the above technique?

seanhess:
Is this branch basically a duplicate of #9? Short of wanting to use a different approach that web-view affords, what is the use case where this is needed? You mentioned figuring it out in another thread.

kfigiela:
This branch addresses point 1 from #9. I opened #9 to discuss hyperbole design and future development in broad sense. As far as my use-case, at the moment I'm running off fork with commits from this branch included. I will investigate that further, but I think it's doable. 

Apart from my use-case, it just feels a principled design to have separation of concerns with hyperbole handling routing/hyperview handling and being agnostic to any styling. This is the only place hyperbole adds a modifier that can't be customized and that was actually frustrating when I found that – I was expecting an extra wrapper element, but not with any styles out there.

seanhess:
It's only there so that views that need to fill their parent can do so. `col` does this already when in a parent container. I definitely think the hyperview should have that property, and I really like keeping `hyper` focused on attaching views, rather than styling.

kfigiela:
How about leaving `flexCol` as a default mod? 

BTW, `onLoad` adds similar extra element with no classes and no way to customize either, so that's a possible todo. 

I did further investigation on my use-case and I'm pretty much sure that I need to be able to pass some styles to either of those elements. I'm doing full-screen dashboard with certain elements scrollable. Some extra CSS is needed for hyper/onLoad extra wrappers so height distributes correctly. Anyway, I'm leaving it here, as I've already spent too much time on that, and there is no issue that inline `<style>` could not fix ;)

seanhess:
Sounds good. When you find an example where you need to pass styles to those elements will you send it to me?

seanhess:
Closing for now, but will reopen once we have some specific examples that can't be handled with the current method

Issue #9, opened 2024-06-10
Hyperbole not to force CSS paradigm


kfigiela:
Opening this in a response to [the comment](https://github.com/seanhess/hyperbole/pull/6#discussion_r1632347227).

[web-view](https://github.com/seanhess/web-view)'s tagline says "write Haskell instead of CSS". This is quite interesting idea and I pretty much like it for ad-hoc styling – especially the dynamic (content-dependent) part. However, I'm not fully sold to use this as the primary way of doing styling. In the real world this may not be always optimal, for instance:

* you decide to use some a CSS library such as Bootstrap (or whatever is trendy at the moment) as your starting point,
* you feel more comfortable writing CSS than learning yet another abstraction layer, or possibly you have dedicated CSS wizard in your team to do the CSS,
* you use some obscure CSS features that the type-safe wrapper does not have abstractions for,
* it really depends on the style of work you and your team do,
* CSS support in code editors and browser dev tools is great giving instant feedback.

I've been playing with hyperbole for a few days now doing some prototyping work for internal tool I work on at work. What I ended up in the end is a mix of:

* CSS library as a starting point
* web-view Haskell CSS for ad-hoc styling,
* regular CSS.

There were a few friction points:

* class name collisions on `row` and `col`,
* hyper adding `col`, being unable to pass mods to it.
* browser dev tools experience was not that great as CSS rules were rendered as a part of dynamic content – overwritten in my case every 1s because of polling,
* lot of this was just `prop` with text-based values, as web-view support is not yet there.

That all being said, I propose the following:

* hyperbole core remains CSS agnostic – enables user to work whatever style and paradigm he wants,
  * essentially – it does not bring any CSS rules on it's own – it could bring some classes (with appropriate prefix to avoid collisions), 
* hyperbole API provides an `Mod` argument whenever applicable to pass extra attributes/styles without hardcoding them (that specifically applies to forms),
* web-view continues to improve support for "write Haskell instead of CSS" by extending CSS features coverage,
* web-view should prefix all classes to avoid collisions – up for the debate if this should only apply to class names hardcoded in web-view library or to all that are defined with it? 
* web-view provides a convenience `Mod` to easily add "external" class.

As for the first item, the only annoyance I found was is `hyper` that adds `col` class. For that one I opened [a pull request](https://github.com/seanhess/hyperbole/pull/8).

seanhess:
[...]

PR #10, opened 2024-06-11
Don't add duplicate CSS rules

kfigiela:
As discussed in #9, there are some glitches related to polling and CSS handling. Essentially, hyperbole will add all CSS rules it gets from hyper view whenever hyper is reloaded. When polling this causes CSS rules list to grow indefinitely. I wonder if that could cause performance issues for long running app? Furthermore, this made dev-tools overrides virtually impossible due to constant updates. To summarize the issue is best described with a video:

b33b5e4a-c464-4973-b691-e797e1da3943

In this PR I attempt to address the issue, by ensuring hyperbole will not add the same rule twice. This deduplicates on what would be added, not necessarily on what is currently in the stylesheet, allowing for local overrides.


seanhess:
Love it! 

Interesting that you could use it to specify local overrides. I think it would make more sense to override it by creating new definitions in Haskell. You're still leaning towards the traditional method of customization by swapping out stylesheets instead of using your programming language to redefine styles (and getting all the type-checking goodness). web-view is a full departure from that method of reuse.

kfigiela:
I use local overrides in dev-tools for experimentation. Feedback loop is instant, which cannot be said to styles encoded in Haskell. The whole recompile and restart cycle is just not efficient.

seanhess:
Ah, thanks for explaining!

PR #12, opened 2024-06-13
Forms typed validate

seanhess:
The Forms in 0.3 didn't support a "Known to be Valid" state, which is required for my app. This design is common in signup forms, when you confirm that a username is available, and the field turns green. 

This branch expands the `Validated` options into the following. It also supports type-based indexing on the Validation type, using type-level lists. So forms now know which fields they contain, based on that type-level list, and the Validation must match. See [Example.Forms](https://github.com/seanhess/hyperbole/blob/fc427e2ab8a529c12231ead2a3bb9505acc8cf44/example/Example/Forms.hs)

```
data Validated fs a = Invalid Text | NotInvalid | Valid
  deriving (Show)

newtype Validation (fs :: [Type]) = Validation [(Text, Validated fs ())]
  deriving newtype (Semigroup, Monoid, Show)
```

I also changed the `Mod` for `field` to `(Validated fs a -> Mod)`, which makes it easy to customize the style for the entire field at once based on the state. It makes it slightly less easy to set default styles on the field, but those can also be set on a nested container or the input itself. Accessing the `Validated fs a` for the field otherwise requires calling `fieldValid`. 

@kfigiela Have you played with forms yet? I'd love some feedback: I'm unsure about the design. It supports many use-cases, but I wish that the validation states were customizable / composable (not easy to do without complicating the types significantly).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Create fully interactive HTML applications with type-safe serverside Haskell. In
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeFamilies #-}
 
+import Data.Text (Text)
 import Web.Hyperbole
 
 main = do
@@ -29,10 +30,10 @@ mainPage = do
 
 
 data Message = Message Int
-  deriving (Generic, Param)
+  deriving (Generic, ViewId)
 
 data MessageAction = Louder Text
-  deriving (Generic, Param)
+  deriving (Generic, ViewAction)
 
 instance HyperView Message where
   type Action Message = MessageAction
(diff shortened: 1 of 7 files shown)


seanhess:
Pushed an example that makes Validated generic, theoretically allowing the user to specify their own validated type

seanhess:
There's an alternative approach, where instead of each field indexing off a different type, we use a record:

```
data UserForm f = UserForm
  { user :: Field f User
  , age :: Field f Int
  }
  
type family Field (f :: Type -> Type)
type instance Field Identity a = a
type instance Field Labels a = Text


```

Then you can customize the type the form is dealing with using that Field type family. So `UserForm Identity` would have values, and `UserForm Labels` would have text labels. 

It's cool, but it's not obvious how to make it easy to carry the field names around (ideally based on selector fields), and also give the user the custom data/validation/whatever they want

PR #19, opened 2024-07-19
Lint, format and enable pre-commit checks

cgeorgii:
What it says in the title :)

### DIFF
--- a/fourmolu.yaml
+++ b/fourmolu.yaml
@@ -1,5 +1,5 @@
 # # Number of spaces per indentation step
-# indentation: 4
+indentation: 2
 #
 # # Max line length for automatic line breaking
 # column-limit: none
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,2 +1,9 @@
+- arguments:
+    - -XOverloadedRecordDot
+
 - ignore: {name: "Use <$>"}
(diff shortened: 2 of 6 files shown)


fourmolu.yaml
```diff
@@ -1,5 +1,5 @@
 # # Number of spaces per indentation step
-# indentation: 4
+indentation: 2
```
cgeorgii, line 2: Note that this was 4 before but the entire codebase is formatted using 2, so I changed it to reflect the actual situation.

seanhess:
I'm not a fan of pre-commit hooks: they get really annoying during rebases and other git deep dives. I'd rather enforce compliance using CI (github actions), and format on save with my editor. Github keeps changing their Haskell runner though, and I haven't figured out the new version. Do they have a nix runner? Could we use it to run tests? 

But I don't develop using nix, so I have no objection to a pre-commit hook there

cgeorgii:
Hm, I never had an issue during rebase (other than having to manually run the hooks) but I get your pain. Since you don't use `nix`, I'll leave it enabled then.

As for CI, I can look into setting up a workflow with nix if that's cool with you.

seanhess:
Sure, that would be great!

Issue #15, opened 2024-07-15
Triggering action for different hyperview


cgeorgii:
Hello! While playing with the contacts example, I added a form to the page to add a new contact:
![image](https://github.com/user-attachments/assets/d5cb1a93-a4e2-40d4-af56-5dac8e027831)

When this new contact form is submitted, I would like the list of contacts to be automatically reloaded. Is emitting an action from a handler function supported by `hyperbole`? Or is there another more idiomatic way to achieve the same effect?

Thanks!

seanhess:
Great question! Is your form a separate HyperView? Or is inside `Contacts`?

If it's part of the `Contacts` HyperView, this can be done by copying the functionality from the `Reload (Maybe Filter)` action. Let's say you've added a new action `AddUser`, as shown below:

```
data ContactsAction
  = Reload (Maybe Filter)
  | Delete Int
  -- My new action
  | AddUser 
  deriving (Show, Read, ViewAction)
  
contacts :: (Hyperbole :> es, Users :> es, Debug :> es) => Contacts -> ContactsAction -> Eff es (View Contacts ())
contacts _ (Reload mf) = do
  us <- usersAll
  pure $ allContactsView mf us
contacts _ (Delete uid) = do
  userDelete uid
  us <- usersAll
  pure $ allContactsView Nothing us
 contacts _ AddUser = do
  -- your custom code 
  readTheFormAndAddTheUser
  
  -- Same as Reload: we've already factored this functionality into easy functions
  us <- usersAll
  pure $ allContactsView Nothing us
```

The principle here is to write view functions that render the contents based on input data. So if your input data change, it's as simple as using the same function to generate the view with updated info.

seanhess:
I just updated [the contacts example](https://github.com/seanhess/hyperbole/blob/main/example/Example/Contacts.hs) to show how this could be done while reusing the form

seanhess:
On the other hand, we need to do something different if the add user form is outside of the `Contacts` view for some reason (for example: if you only wanted to show it after they press a button, you could put it in its own HyperView).

Let's say you created a new HyperView called NewConcat for this:

```
data NewConcat = NewConcat
  deriving (Show, Read, ViewId)

data NewContactAction
  = OpenForm
  | AddContact
  deriving (Show, Read, ViewAction)
  
 instance HyperView NewContact where
  type Action NewContact = NewContactAction

newContact :: NewContact -> NewContactAction -> Eff es (View NewContact ())
[...]

PR #22, opened 2024-07-23
Extract functionality into their own modules

cgeorgii:
Hi @seanhess, while trying to get an overview of the code before taking a stab at implementing [server pushing](https://github.com/seanhess/hyperbole/issues/15#issuecomment-2239683567) I ended up moving code around a bit. I tried my best to not change any functionality at all but I'm not sure if this might've affected Haddock docs? Probably not. :thinking: 

Some notes:
- I created a dedicated module for the `Server` effect (in `Web.Hyperbole.Effect.Server`), so a lot of the code previously in `Web.Hyperbole.Application` ended up there.
- Dedicated modules for `Request`, `Response`, `Event`, `Host` types.
- Small readability changes (variable names, function ordering)

I understand that this is code churn and I mostly did it to get a deeper understanding of what's going on. I personally think this will allow the code to evolve better but I know you have ongoing efforts that might be impacted by merge conflicts after this, so feel free to ignore.

### DIFF
--- a/example/hyperbole-examples.cabal
+++ b/example/hyperbole-examples.cabal
@@ -27,9 +27,14 @@ executable examples
       Web.Hyperbole
       Web.Hyperbole.Application
       Web.Hyperbole.Effect
+      Web.Hyperbole.Effect.Server
       Web.Hyperbole.Embed
+      Web.Hyperbole.Event
       Web.Hyperbole.Forms
+      Web.Hyperbole.Host
       Web.Hyperbole.HyperView
+      Web.Hyperbole.Request
+      Web.Hyperbole.Response
       Web.Hyperbole.Route
       Web.Hyperbole.Session
       Web.Hyperbole.View
--- a/hyperbole.cabal
+++ b/hyperbole.cabal
@@ -31,9 +31,14 @@ library
       Web.Hyperbole
       Web.Hyperbole.Application
       Web.Hyperbole.Effect
+      Web.Hyperbole.Effect.Server
       Web.Hyperbole.Embed
+      Web.Hyperbole.Event
       Web.Hyperbole.Forms
+      Web.Hyperbole.Host
       Web.Hyperbole.HyperView
+      Web.Hyperbole.Request
+      Web.Hyperbole.Response
       Web.Hyperbole.Route
       Web.Hyperbole.Session
       Web.Hyperbole.View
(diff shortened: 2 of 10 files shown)


seanhess:
The codebase definitely needs some refactoring, but I'd like to wait until we merge (or reject) the two big PRs that are open for feedback. 

That said, let's discuss the approach. Why do you prefer such small modules for each type? It doesn't seem to be very common in libraries, except when those modules might need to be imported separately. Some examples:

* [Scotty](https://hackage.haskell.org/package/scotty-0.22) has 4 modules + its base. Each module exists so people can import optional functionality, like the Trans or Cookie
* [Wai](https://hackage.haskell.org/package/wai) - just one main module and an Internal module. 
* [Yesod](https://hackage.haskell.org/package/yesod) - A huge framework, looks like these guys split things up into separate packages, and those packages are pretty granular

 For hyperbole, I was going for a logical grouping of related functionality and types. It's gotten out of hand, but I would rather not have too long a list of modules. I think that can feel intimidating to users? 
 
 Closing for now, but happy to continue discussing (or open an issue to talk about it).

cgeorgii:
> I think that can feel intimidating to users?

It's possible to only expose certain modules to users and re-export types from a couple of top-level modules, if that's the main concern (using `exposed-modules`/`other-modules` in the package.yaml file. Even local documentation can be re-exported this way, I believe.

I think in general having small, focused modules is a good idea. It helps with testing and exposes concepts that are "wrong" or that could be simplified.  

For instance, during this refactor it became apparent that the `SocketError` is more related to parsing requests than sockets since it only has one constructor for that case and moving it into the server effect module and away from the parsing code would make little sense. Another example is the `Event` type, that could probably have just one type parameter, if we slightly change the code that uses it.

Of course it's possible to reach the same conclusions with the current code organization, but I find finer-grained modules better for understanding exactly what you need and not more, which is important with unfamiliar code bases.

Finally, I hope hyperbole becomes the next best thing, so having a highly organized codebase might help with avoiding conflicts and keeping changes focused when the inevitable billions of pull requests come crashing through 😝

seanhess:
@cgeorgii I just pushed a big refactor to move things into more specific modules. It's not quite as aggressive as what you had in this PR, but I think it's logically organized. What do you think? Thanks for prompting this!

PR #37, opened 2024-10-28
fix for ghc 9.10.1

benjaminweb:


seanhess:
Weird, why does `Data.Text` export show?

Issue #31, opened 2024-10-16
Simplify handle / load interface design


benjaminweb:
Current handle load interface goes like

` handle central $ handle presets $ handle handler3 $ load $ do`

A handler's function is to catch any actions originating from any views.
`handle` currently wraps a handler around other handlers so the actions get caught.
A handler will not work

Current architecture:
- function type like `'[View1, View2, View3]` will let the body only compile if the body wraps load in all handlers
- this is not intuitive
- problem is that handlers can't be a list because of heterogeneous type

Desirable qualities of new `handle / load` interface design:
- composition of handlers mirroring the actual composition of views => more intuitive
- _… what else?_

seanhess:
What design would be an improvement on what we have now?

Here are good options that aren't possible:
1. Lists (not heterogenous)
2. Monadic interface, like in the previous version (same problem, all the types need to be the same. can't build one)

Other Options I considered

*Operators* - we could create operators that mimic what handle and load are doing. Something like this (Using Example.Contacts)

```
-- I'm not saying these are the best operators, just showing what might be possible. 
page :: (Hyperbole :> es, Users :> es, Debug :> es) => Page es '[Contacts, Contact]
page = do
 contacts :|: handle contact :$: do
    ...
```

*Type Operators* - Similar to Servant, use type operators instead of a type-level list. Then use the same operators to link the handlers together. 

```
page :: (Hyperbole :> es, Users :> es, Debug :> es) => Page es (Contacts :|: Contact)
page = do
 contacts :|: handle contact :$: do
    ...
```

*Tuples* - We could make use tuples instead? They get unweidly with a bunch of handlers though! But that could be solved with some sort of factoring / nested tuples? 

```
-- we could keep the type-level list, or a use a type-level tuple
page :: (Hyperbole :> es, Users :> es, Debug :> es) => Page es (Contacts, Contact)
page = do
 handle (contacts, contact) $ do 
  ...
```

Have any other ideas?

seanhess:
Good question. If they are handled explicitly, you can easily pass app context to a handler, or more importantly, to reuse hyperviews

Context - as in a db connection or something. This isn't that useful, since it's easy to use the `Reader` to accomplish the same thing

```
myHandler :: Connection -> MyView -> MyAction -> Eff es (View MyView ())
myHandler conn _ Save = Db.saveSomething "hello"
```

[...]

PR #43, opened 2024-11-05
Class handlers new

seanhess:
I managed to finally figure out how to resolve handlers via a typeclass. This is an alternative to #41, and in response to #31. 

Unfortunately we couldn't fit it into the `HyperView` class, because this class needs to carry the `es :: [Effect]` type variable. Multiple functions require that `HyperView` work outside of effect handlers, like `button :: HyperView id => Action id -> Mod -> View id ()`

Here's what they look like in this branch:

```
{-# LANGUAGE UndecidableInstances #-}

page :: (Hyperbole :> es, Concurrent :> es, Reader (TVar Int) :> es) => Page es '[Counter]
page = load $ do
  var <- ask
  n <- readTVarIO var
  ...

instance HyperView Counter where
  type Action Counter = Count

instance (Reader (TVar Int) :> es, Concurrent :> es) => Handle Counter es where
  handle _ Increment = do
    n <- modify $ \n -> n + 1
    pure $ viewCount n
```

They also require UndecidableInstances. But other than those two downsides, the advantages of being able to look up a handler by hyperview are great. That page function is much simpler.

### DIFF
--- a/example/Example/AppRoute.hs
+++ b/example/Example/AppRoute.hs
@@ -1,0 +1,40 @@
+module Example.AppRoute where
+
+import Data.Text (Text)
+import Example.Effects.Users (UserId)
+import Web.Hyperbole
+
+
+data AppRoute
+  = Main
+  | Simple
+  | Hello Hello
+  | Contacts ContactRoute
+  | Transitions
+  | Query
+  | Counter
+  | Forms
+  | LiveSearch
+  | Sessions
+  | Redirects
+  | RedirectNow
+  | LazyLoading
+  | Concurrent
+  | Errors
+  deriving (Eq, Generic)
+instance Route AppRoute where
+  baseRoute = Just Main
+
+
+data ContactRoute
+  = ContactsAll
+  | Contact UserId
+  deriving (Eq, Generic)
+instance Route ContactRoute where
+  baseRoute = Just ContactsAll
+
+
+data Hello
+  = Greet Text
+  | Redirected
+  deriving (Eq, Generic, Route)
--- a/example/Example/Concurrent.hs
+++ b/example/Example/Concurrent.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE UndecidableInstances #-}
+
 module Example.Concurrent where
 
 import Data.Text (pack)
@@ -6,9 +8,9 @@ import Example.Effects.Debug
 import Web.Hyperbole
 
 
-page :: (Hyperbole :> es, Debug :> es, IOE :> es) => Page es Contents
+page :: (Hyperbole :> es, Debug :> es, IOE :> es) => Page es '[Contents]
 page = do
-  handle content $ do
+  load $ do
     pure $ do
       col (pad 20) $ do
         hyper (Contents 50) $ viewPoll 1
@@ -26,13 +28,11 @@ data ContentsAction
 
 instance HyperView Contents where
   type Action Contents = ContentsAction
-
-
-content :: (Hyperbole :> es, Debug :> es, IOE :> es) => Contents -> ContentsAction -> Eff es (View Contents ())
-content (Contents dl) (Load n) = do
-  -- BUG: this is blocking the thread, so the short poll has to wait for the long to finish before continuing
-  delay dl
-  pure $ viewPoll (n + 1)
+instance (Debug :> es) => Handle Contents es where
+  handle (Contents dl) (Load n) = do
+    -- BUG: this is blocking the thread, so the short poll has to wait for the long to finish before continuing
+    delay dl
+    pure $ viewPoll (n + 1)
 
 
 viewPoll :: Int -> View Contents ()
--- a/example/Example/Concurrent.hs
+++ b/example/Example/Concurrent.hs
@@ -8,7 +8,7 @@ import Example.Effects.Debug
 import Web.Hyperbole
 
 
-page :: (Hyperbole :> es, Debug :> es, IOE :> es) => Eff es (Page '[Contents])
+page :: (Hyperbole :> es, Debug :> es, IOE :> es) => Page es '[Contents]
 page = do
   pure $ do
     col (pad 20) $ do
@@ -28,7 +28,8 @@ data ContentsAction
 instance HyperView Contents where
   type Action Contents = ContentsAction
 instance (Debug :> es) => Handle Contents es where
-  handle (Contents dl) (Load n) = do
+  handle (Load n) = do
+    Contents dl <- viewId
     -- BUG: this is blocking the thread, so the short poll has to wait for the long to finish before continuing
     delay dl
     pure $ viewPoll (n + 1)
--- a/example/Example/Concurrent.hs
+++ b/example/Example/Concurrent.hs
@@ -8,13 +8,12 @@ import Example.Effects.Debug
 import Web.Hyperbole
 
 
-page :: (Hyperbole :> es, Debug :> es, IOE :> es) => Page es '[Contents]
+page :: (Hyperbole :> es, Debug :> es, IOE :> es) => Eff es (Page '[Contents])
 page = do
-  load $ do
-    pure $ do
-      col (pad 20) $ do
-        hyper (Contents 50) $ viewPoll 1
-        hyper (Contents 1000) $ viewPoll 100
+  pure $ do
+    col (pad 20) $ do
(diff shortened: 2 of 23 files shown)


seanhess:
@benjaminweb can I get your feedback on this approach vs #41? Which do you think is better?

benjaminweb:
Will do once I’ve tried out type class and tuple approach (days)!

seanhess:
One other downside to this approach: it makes passing arguments to handlers more complicated. It's not that big of a deal to add a `Reader whatever :> es` constraint and runReader in the main app, but ghc is resisting my attempts to make it casual to add context when defining the page:

Before:

```
page :: (Hyperbole :> es, Concurrent :> es) => TVar Int ->  Page es '[Counter]
page var = handle (counter var) $ do
  n <- readTVarIO var
  ...
```

Now:

```
-- you MUST call runReader from outside of this function
-- if you attempt to runReader inside it, the types don't line up (because the es for the page has to match)
page :: (Hyperbole :> es, Concurrent :> es, Reader (TVar Int) :> es) =>  Page es '[Counter]
page = do
  -- can't do this here: runReader var $ do
  n <- readTVarIO var
  ...
```

This isn't a problem for something like Counter, because the top-level app controls that TVar anyway, but it is a problem in my real application at work, where I've created a reusable hyperview with a cancel button that can does different things on two pages. The handler for It looks like this (using the syntax of #41)

```
inversions :: (Hyperbole :> es, ...) => Eff es (View InversionStatus ()) -> InversionStatus -> InversionAction -> Eff es (View InversionStatus ())
inversions onCancel (InversionStatus ip iip ii) = \case
  Cancel -> do
    send $ Inversions.Remove ii
    onCancel
  Download -> do
    ...
```

It is used on two different pages with a different `onCancel`. 

```
page1
  :: (Hyperbole :> es, ...)
  => Id Proposal
  -> Id Inversion
  -> Eff es (Page (InversionStatus : Require InversionStatus))
page1 ip i = handle (inversions redirectHome, ...) $ do

page
  :: (Hyperbole :> es, ...)
  => Id Proposal
  -> Id InstrumentProgram
  -> Eff es (Page (ProgramInversions : ProgramDatasets : InversionStatus : InversionViews))
page ip iip = do
  handle (inversions (deleteInversion ip iip), ...) $ do
```

Maybe reusing a hyperview like this is a bad idea. In which case this being a PITA is a good thing?

benjaminweb:
Incorporated both variants into my playground:
#41: https://github.com/benjaminweb/hyperbole-tutorial/blob/tuple-list-handlers/app/Main.hs
#43:https://github.com/benjaminweb/hyperbole-tutorial/blob/typeclass-list-handlers/app/Main.hs

I instantly like #43 more.
It is appealing to simplify the interfacing (and/through documented examples) further. That's my intuition right now. — What's your's?

Observations:
- #43 makes it to me much clearer what the purpose of the `handle` function is, point for #43
- #41 requires less changes to existing code

benjaminweb:
Regarding Passing context to handlers.
In my (limited) experience I've passed to handlers mostly the same arguments (contexts).
I wished there was / I knew a better way to pass it around.

I haven't encountered the situation of parametrized handlers for reusing hyperviews yet.

Naive question: What you've done for #43, namely engineering the variants of handlers (handled by load?) on the type class level, could you do similar for the arguments of handlers? Though, it might be overkill or not viable at all, just sharing this notion with you.

seanhess:
Thanks, that’s very helpful. I had been leaning towards #41 because it is more explicit, but I think you are right. A lot of the design comes from Elm, but that throws off my intuition sometimes because Elm doesn’t have type classes. I think you’re right that #43 will feel more intuitive to Haskell programmers. 

> https://github.com/seanhess/hyperbole/pull/43 makes it to me much clearer what the purpose of the handle function is

> Naive question: What you've done for https://github.com/seanhess/hyperbole/pull/43, namely engineering the variants of handlers (handled by load?) on the type class level, could you do similar for the arguments of handlers? Though, it might be overkill or not viable at all, just sharing this notion with you.

These comments taken together are interesting. We pass the View Action into the handler so the user can pattern match on it. It’s the simplest way to make sure they handle all possible actions. 

But the View Id is only useful for context: database ids for example (see Example.Contacts). The view Id could certainly be moved into a `Reader id :> es` constraint. That would be much cleaner in the common use case where a handler doesn’t need the view id for anything. 

We could theoretically do something similar for the action, but can you imagine a different way to specify what an action does?

seanhess:
I just pushed a new update to this branch that incorporates your idea of hiding the view id behind a reader context. I like it! 

Here's what it looks like if you don't need the view id: Example.Counter

```
instance HyperView Counter where
  type Action Counter = Count
instance (Reader (TVar Int) :> es, Concurrent :> es) => Handle Counter es where
  handle Increment = do
    n <- modify (+ 1)
    pure $ viewCount n
  handle Decrement = do
    n <- modify (subtract 1)
    pure $ viewCount n
```

If you DO need the view id, there's a new funciton `viewId :: Reader id :> es => Eff es id`

```
instance HyperView Contact where
  type Action Contact = ContactAction
instance (Users :> es, Debug :> es) => Handle Contact es where
  handle action = do
    -- No matter which action we are performing, let's look up the user to make sure it exists
    Contact uid <- viewId
    u <- Users.find uid
    case action of
      View -> do
        pure $ contactView u
      Edit -> do
        pure $ contactEdit u
      Save -> do
        delay 1000
        unew <- parseUser uid
        Users.save unew
        pure $ contactView unew
```

To put it through its paces, and to make sure that we have alternatives to argument-passing for reuse, I updated the advanced Contacts example to demonstrate reusing a HyperView on different pages. 

The new Example.Contact lets a user edit a contact from a separate page:

```
-- Example adding a reader context to the page, based on an argument from the AppRoute
response :: (Hyperbole :> es, Users :> es, Debug :> es) => UserId -> Eff es Response
response uid = runReader uid $ runPage page


-- The page assumes all effects have been added
page
  :: forall es
   . (Hyperbole :> es, Users :> es, Debug :> es, Reader UserId :> es)
  => Page es '[Contact]
page = do
  uid <- ask
  u <- Users.find uid
  pure $ do
    col (pad 10 . gap 10) $ do
      hyper (Contact uid) $ contactView u

```

Example.Contacts uses the `Contact` view from Example.Contacts, modifying what the edit button does to add a delete button

```
-- Reuse Contact View ----------------------------------

-- Make a newtype so we can customize the behavior of the handler
newtype InlineContact = InlineContact Contact
  deriving newtype (ViewId)


instance HyperView InlineContact where
  type Action InlineContact = ContactAction
instance (Users :> es, Debug :> es) => Handle InlineContact es where
  handle Edit = do
    -- Edit will show the normal view, plus a delete button
    InlineContact (Contact uid) <- viewId
    u <- Users.find uid
    pure $ inlineEdit u
  handle other = do
    -- all the other views are handled by `Handle Contact es`
    InlineContact c <- viewId
    asdf <- runReader c $ handle @Contact other
    pure $ addContext c asdf

inlineEdit :: User -> View InlineContact ()
inlineEdit u = onRequest contactLoading $ col (gap 10) $ do
  InlineContact c <- context
  addContext c $ contactEdit' u
  target Contacts $ button (DeleteUser u.id) (Style.btn' Danger . pad (XY 10 0)) (text "Delete")

```

benjaminweb:
That's pretty straightforward: https://github.com/benjaminweb/hyperbole-tutorial/commit/ac46e4823ca30063073c09ed5ecd87429c70a836
It's simpler because we don't use `load` any more.

…looking at: https://github.com/seanhess/hyperbole/blob/6f36420cd8770a316fa9546589131892de44caae/example/HelloWorld.hs
Would it be simpler to define a View via a type instance?

seanhess:
> …looking at: https://github.com/seanhess/hyperbole/blob/6f36420cd8770a316fa9546589131892de44caae/example/HelloWorld.hs Would it be simpler to define a View via a type instance?

If you mean adding a view function to the Handler or HyperView class, no I don't think that would help. Views need data to render, like a type from the database or something. Right now, you can gather anything you need in a handler and pass it directly to the view. To add a view function to one of the typeclasses we would also have to add some "model" datatype, which would complicate it more than I'm comfortable with.

seanhess:
I just switched my work project over to this branch and it's working great. Let's pull the trigger. Thanks for the feedback @benjaminweb!

PR #44, opened 2024-11-07
Client-side triggers. Resolves #15

### LINKED ISSUE #15: Triggering action for different hyperview

seanhess:
This PR adds support for triggers. Resolves #15 

If you know that another hyperview exists, you can trigger 

```
controls :: (Hyperbole :> es) => Controls -> ControlsAction -> Eff es (View Controls ())
controls _ (Broadcast t) = do
  trigger (Message 0) (Say t)
  trigger (Message 1) (Say t)
  trigger (Message 2) (Say t)
  pure viewControls
```

TODO

- [x] Document Example.Triggers to show how this is more easily done WITHOUT triggers
- [x] Type-check that Message is handled in Require?


seanhess:
I can't figure out how to make this type-safe: you can attempt to trigger any `HyperView`, even if it isn't handled or doesn't exist:

```
controls :: (Hyperbole :> es) => Controls -> ControlsAction -> Eff es (View Controls ())
controls _ (Broadcast t) = do
  trigger (Message 0) (Say t)
  trigger (Message 1) (Say t)
  trigger (Message 2) (Say t)
  trigger NotHandledView OhNo
  pure viewControls
```

And all we get is a lousy javascript console error :(

To fix, we would need to keep the hyperview's type in the effect somehow, because trigger doesn't know which view's handler you are resolving. 

```
trigger :: (Hyperbole :> es, HyperView id) => id -> Action id -> Eff es ()
```

If we adopt #43, we can automatically handle it, but there's no way to know the view exists. 

I'm unsure if this is a good idea. I know people are going to use/abuse it if included, and perhaps there's always a better way to handle this. I guess `target` has the same problem. You're guessing that the other view exists, but we can at least enforce that it is handled. 

One option would be to make this something that goes in the view instead,

seanhess:
With the class-based handlers in #43, there's a `(Reader id : es)` in the handler type. That means we should be able to make triggers type-safe. Update to come next week

cgeorgii:
Hey @seanhess! Thanks for pinging me. I've been swamped lately but I'll try to take a look and let you know what I think on Friday, alright? :)

seanhess:
That would be excellent, thank you!

cgeorgii:
Hey @seanhess, after playing around I'm pretty convinced by the nested approach! I can't think of a reason to use the other, to be honest. Thanks for the effort, I'll play around a bit more and let you know if anything comes up :)

seanhess:
Great, thanks for your feedback. 

Closing this, but will reopen if anyone can think of a motivating example for trigger

cgeorgii:
Hey @seanhess, I've been thinking of a use-case that I wouldn't know how to model without triggers.

Imagine that in the [Example.Triggers.Nested](https://github.com/seanhess/hyperbole/blob/trigger-actions-15/example/Example/Triggers/Nested.hs), example I wanted to add a new button to the main view that changes an aspect, e.g some background color, of a specific listener without changing the current string being displayed by it. Having in mind that the listener can change this string themselves by clicking "Say Hi", could it be done using Hyperbole?

In summary:
- A button labeled "Choose one" similar to "Clear", "Broadcast hello", etc.
- Below "Broadcast goodbye"
- When clicked, selects one of the listeners (randomly) and changes its background color while keeping the text previously displayed unchanged.

image

seanhess:
If you need to update a specific HyperView, and for it to maintain its current state, it's probably time to add some sort of real state to it. HyperViews have a sort of implicit state if you squint, of whatever was passed into the funciton that rendered them, but it's not very powerful. 

It can be done with a TVar, Sessions, or a database. 

There might be an argument for automatically supporting state in views, similar to your PR #47, but let's debate that in the other thread.

PR #47, opened 2024-12-01
Component API

cgeorgii:
Hey @seanhess, I am pretty excited to share a new experimental way of writing apps using `hyperbole`. I think it might be easier for newcomers since it groups everything one needs in one convenient typeclass called `Component`.

Without further ado, here's how the `Simple.hs` example looks like using this new approach:

```hs
page :: (Hyperbole :> es, Concurrent :> es, Reader (TVar Int) :> es) => Page es '[Counter]
page = do
  var <- ask
  n <- readTVarIO var
  pure $ col (pad 20 . gap 10) $ do
    el id "Counter"
    start Counter CounterModel{count = n}

data Counter = Counter
  deriving (Show, Read, ViewId)

instance Component Counter es where
  data Model Counter = CounterModel
    { count :: Int
    }

  data Action Counter
    = Increment
    | Decrement
    deriving (Show, Read, IsAction)

  -- Add additional effects required by the component. Can be omitted if there are none.
  type Effects Counter es = (Reader (TVar Int) :> es, Concurrent :> es)

  render :: Model Counter -> View Counter ()
  render model = col (gap 10) $ do
    row id $ do
      el (bold . fontSize 48 . border 1 . pad (XY 20 0)) $ text $ pack $ show model.count
    row (gap 10) $ do
      button Decrement id "Decrement"
      button Increment id "Increment"


  -- Ideally the update function would only run effects and update the model without having to render the view manually. Its signature could change to receive the previous model and the action instead of just the action.
  update = \case
    Increment -> do
      -- Inside the update function we have access to the effects defined in the
      -- associated type 'Effects c es':
      n <- modify $ \n -> n + 1
      pure $ render $ CounterModel{count = n}
    Decrement -> do
      n <- modify $ \n -> n - 1
      pure $ render $ CounterModel{count = n}

```

The class itself looks like this:

```hs
-- | A 'Component' is a self-contained piece of a 'Page'.
class (ViewId c, IsAction (Action c)) => Component c es where
  -- | The data used by the component.
  data Model c

  -- | The actions supported by the component.
  data Action c

  -- | Additional effects required by the component. Can be omitted.
  type Effects c (es :: [Effect]) :: Constraint
  type Effects c es = ()

  -- | Other components nested in the component. Can be omitted.
  type Require c :: [Type]
  type Require c = '[]

  -- | Render the component.
  render :: Model c -> View c ()

  -- | Update the component based on a message. Ideally would only change the data model and leave rendering to the 'render' function.
  update :: (Effects c es) => Action c -> Eff (Reader c : es) (View c ())
```
In essence, we now need to declare the component id like before, deriving `ViewId` and giving it a `Component` instance. The user no longer needs to worry about `HyperView` and `Handler` classes.

### More examples
I added a [`Nested.hs`](https://github.com/seanhess/hyperbole/pull/47/files#diff-c611a4ea41f8579edcdc7cedce5e66f2935810e7c25f9c30d82826bebb4c020b) file next to the `Simple.hs` example to showcase the nested approach using `Require`. 

I also have a repository using this version that reimplements all examples using this new style (well, a couple of them - still working on the others) that I can share with you once it's in a more ready state.

### Issues/TODOs
- I couldn't get the tests to run locally but I can work on adjusting them (and the examples) if you think this is a good approach
- I couldn't figure out how to make `delegate` function work (but also didn't spend so much time on it) :shrug:.

Anyway, I'd love to hear your thoughts about this!

### DIFF
--- a/example/hyperbole-examples.cabal
+++ b/example/hyperbole-examples.cabal
@@ -1,6 +1,6 @@
 cabal-version: 2.2
 
--- This file has been generated from package.yaml by hpack version 0.37.0.
+-- This file has been generated from package.yaml by hpack version 0.36.1.
 --
 -- see: https://github.com/sol/hpack
 
@@ -23,9 +23,11 @@ source-repository head
 executable examples
   main-is: Main.hs
   other-modules:
+      Nested
       Simple
       Web.Hyperbole
       Web.Hyperbole.Application
+      Web.Hyperbole.Component
       Web.Hyperbole.Effect.Hyperbole
       Web.Hyperbole.Effect.Server
       Web.Hyperbole.Embed
--- a/hyperbole.cabal
+++ b/hyperbole.cabal
@@ -1,6 +1,6 @@
 cabal-version: 2.2
 
--- This file has been generated from package.yaml by hpack version 0.37.0.
+-- This file has been generated from package.yaml by hpack version 0.36.1.
 --
 -- see: https://github.com/sol/hpack
 
@@ -31,9 +31,11 @@ source-repository head
 
 library
   exposed-modules:
+      Nested
       Simple
       Web.Hyperbole
       Web.Hyperbole.Application
+      Web.Hyperbole.Component
       Web.Hyperbole.Effect.Hyperbole
       Web.Hyperbole.Effect.Server
       Web.Hyperbole.Embed
--- a/src/Nested.hs
+++ b/src/Nested.hs
@@ -22,7 +22,7 @@ page = do
 
 
 data MainView = MainView
-  deriving (Show, Read, ViewId, HyperView)
+  deriving (Show, Read, ViewId)
 
 
 instance Component MainView where
@@ -66,7 +66,7 @@ mainControls = col (gap 5) $ do
 
 
 data Listener = Listener Int
-  deriving (Show, Read, ViewId, HyperView)
+  deriving (Show, Read, ViewId)
 
 
 instance Handle Listener es where
--- a/src/Nested.hs
+++ b/src/Nested.hs
@@ -25,7 +25,7 @@ data MainView = MainView
 
 
 instance Component MainView es where
-  data Msg MainView
+  data Action MainView
     = Broadcast Text
     | ClearAll
     deriving (Show, Read)
@@ -36,7 +36,7 @@ instance Component MainView es where
     }
 
 
-  type Import MainView = '[Listener]
+  type Require MainView = '[Listener]
 
 
   render model = do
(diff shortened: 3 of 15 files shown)


seanhess:
Will you create a separate PR for only the change that removes the `Handle` typeclass, merging it into `HyperView`? We can implement that separately from the design discussion here. I attempted to implement it myself, but ran into issues where the view couldn't figure out `es` at the use site.

seanhess:
Nevermind, I figured it out! We can drop the `es` from `HyperViewHandled` because it mentions `Require id` on the right hand side, which has the same effect as what you figured out for `button` et al. See #49

cgeorgii:
Awesome, thanks for taking the time to write this! It did feel like I was trying to shoehorn the Elm architecture into hyperbole and I wasn't sure I liked the idea either.

I was working on a PR like you asked but I see you already opened two of them. Anything you would like me to do or do you have it covered?

seanhess:
Those other 2 PRs cover the two improvements you came up with. Thanks so much! I LOVE how much cleaner the class definition is with both. Sorry if I stole your thunder, but I was really excited to implement them. 

Does this mean you prefer the current stateless design? I'm happy to talk about it more, I just wanted to make sure you had the full background before continuing. Anything else you think we should use from your idea?

seanhess:
Should we rename `handle` to `update`? The function certainly does "update" the view. Or does that invoke TEA too much?

cgeorgii:
> Those other 2 PRs cover the two improvements you came up with. Thanks so much! I LOVE how much cleaner the class definition is with both. Sorry if I stole your thunder, but I was really excited to implement them.

Hehe, I was quite excited as well but no worries, I should've told you I was on it. It would've been nice to have a first meaningful contribution but it will come.

> Does this mean you prefer the current stateless design?

I'm not sure, but at the same time I haven't really tried building anything real with hyperbole yet. I'm going to take a shot at implementing the use-case I mentioned [here](https://github.com/seanhess/hyperbole/pull/44#issuecomment-2510315052) to see how it goes.

> Anything else you think we should use from your idea?
No, I think that was all of it, pretty much. We removed the associated `Model` and `render` functions but it still is pretty much the `Component` class I wrote, right? 

Ah, there's also the associated `Effects` type. It listed any sort of additional effects that the `HyperView`/`Component` might. Was it superfluous?

cgeorgii:
Closed in favor of #48 and #49.

cgeorgii:
Closed in favor of #48 and #49.

seanhess:
Also, just pushed some refactors, grab the latest before going too deep. Thanks again!

cgeorgii:
> Should we rename `handle` to `update`? The function certainly does "update" the view. Or does that invoke TEA too much?

I think `update` is more straightforward and I'd prefer it but no strong feelings either way.

seanhess:
Let’s do it. I like update more. I’m traveling all week, feel free to make the change if you like!

PR #50, opened 2024-12-14
Fixes flake.nix and sets up nix based ci

Skyfold:
I went ahead and created this pull request since I wanted to ask you about making hyperbole-examples depend on hyperbole instead of refercing the code directly. I know this would likely break your workflow for development, but I wasn't able to figure out a way to make nix work with the `../src` path. Let me know if this is ok or more about your workflow to see if I can find a workaround.

* Switch from Cabal/GHC container to Nix-based build system
* Setup devShell for hyperbole and hyperbole-examples

Changes:
- Replace CI Cabal/GHC container setup
- Add matrix strategy for multi-OS support
- hyperbole-examples.cabal add hyperbole to build-depends (removed ../src from source dirs)
- updated README with flake usage
- applied formolu to Simple.hs so pre-commit-check passes

The flake refrences my fork of web-view, other than that it is ready to merge.


Skyfold:
I'll update this once you have updated hyperbole for web-view 0.6.0.

Skyfold:
The nix setup is similar to `web-view` with the same kind of checks that skip the development shells. It does test that the hyperbole-example code compiles with the hyperbole source.

All that is missing is to change `Skyfold` to `seanhess` for the `web-view` source once it is merged. I'll put out another commit to fix that.

Skyfold:
This is ready to merge

seanhess:
I'm linking the src folder in from examples so that ghcid reloads if changes are made in either the examples directory or src. Do you know of an alternative? If we do it the normal way, and have hyperbole-examples depend on hyperbole, ghcid inside of examples will only reload for changes in the examples folder. Changes in the src directory require restarting. 

It's not a huge deal. I'd like to hope refactors are going to become less common, but it hasn't happened yet. 

I'm ok going ahead with this if we can't think of a way around it

Skyfold:
Requiring ghcid to recompile hyperbole when working on both is not ideal. I would also like to have one development shell for both projects and have `go to definition` work from the examples project for `hyperbole` code.

I realised there is an easy way: put both hyperbole and hyperbole-examples dependencies in the same shell and keep the `../src` as is. 

There is one other approach. We can set `multi-repl: True` in cabal.project and merge the hyperbole-examples.cabal with hyperbole.cabal. The `multi-repl` means that `cabal repl` will load all tests, libraries, executable, into one repl. When you call ghcid with `-c cabal repl` it won't need to restart even if you change the hyperbole source. This would make it convenient to work on all parts of the project from one `cabal repl`.

Let me know which approach you would prefer.

Skyfold:
I wanted to show you what it would look like and get to try out the multi-repl, so I went ahead and combined them. If it doesn't work well, its easy to revert and try the other way.

Skyfold:
I added some code to `example/Main.hs` for hot reloading the example server. I borrowed the idea and code from yesod's template that provides a fast way to reload the server inside of ghci. With the latest commit you can run:

```
ghcid --setup=Main.update --command="cabal repl exe:examples lib:hyperbole test" --run=Main.update --warnings
```

This loads the examples, library and tests in one repl, but targets the `exe:examples` so you can run the `update` function from `example/Main.hs`. Changes do not require a restart, they will be interpreted and if successful, the hot reloading function will restart the example server.

Note: In my projects I have a "refresh firefox" as part of the hot reload, but that depends on my window manager. I have not found a universal way to trigger a browser refresh.

seanhess:
Hot reloading, neat! I'll hopefully get to trying this today. I believe "livereload" is sort-of considered a standard for browser refresh.

seanhess:
The reason why we had a separate package for examples is because hackage ends up requiring the dependencies of example too. Take a look at what your branch looks like as a candidate: https://hackage.haskell.org/package/hyperbole-0.4.2/candidate

See how wai-middleware-static, foreign-store, filepath, and directory are listend as dependencies? It feels sloppy. Do you know of a workaround?

I suppose it's not a big deal if it's only one or two small dependencies. Directory and FilePath are common, so they're probably already compiled somewhere in the tree. wai-middleware-static isn't much added on to wai and warp. 

The hot reload is cool, but I'm worried about making the example too complicated, and making maintenance too brittle. My experience with other frameworks is that it breaks all the time. What do you think?

Skyfold:
I think we can have our cake and eat it too. I'll put the executables back into their own cable file and then add it to the top level cabal.project file (and remove the example one). This should let us still load everything in a single repl, while keeping the examples separate.

seanhess:
Sounds good!

seanhess:
In terms of having my cake, the things I would like to work are:

The `bin/dev` shell command in the repo should run tests, compile the js, and turn on ghcid (or equivalent) which reloads when a source file is changed anywhere in the repository

`haskell-language-server-wrapper` should run on all the files, including the examples and src directory (this is what makes editor support work for me)

Skyfold:
> `haskell-language-server-wrapper` should run on all the files, including the examples and src directory (this is what makes editor support work for me)

This latest commit sets up HLS to work across the entire cabal project (with a separate hyperbole-examples.cabal file).

> The bin/dev shell command in the repo should run tests, compile the js, and turn on ghcid (or equivalent) which reloads when a source file is changed anywhere in the repository

I'm still working on setting up the nodejs part with nix. Once that works, I think your `client()` command should just work inside the nix shell. We can modify the `compile()` command to be one of these two:

If you want to work on both the hyperbole library and example code, this `ghcid` command will run (and hot reload) the examples server as you change any non-testing code.

```
ghcid --setup=Main.update --command="cabal repl exe:examples lib:hyperbole" --run=Main.update --warnings --reload=./client/dist/hyperbole.js
```

If you want to work on the test suite, this will run the tests each time any library code is changed.

```
ghcid --command="cabal repl test lib:hyperbole" --run=Main.main --warnings --reload=./client/dist/hyperbole.js
```

I don't think I can get ghcid to actually run `npx webpack -w`, but it will reload if there are changes.

seanhess:
My only requirement is that bin/dev works outside of nix. I'm not interested in investing the time to figure nix out, so I'll continue to develop with my os and cabal.  

> I don't think I can get ghcid to actually run npx webpack -w, but it will reload if there are changes.

There's no need for this. `npx webpack -w` already watches the js files for changes. 

I'm ready to merge this as soon as ghcid still works and you tell me it's ready.

seanhess:
Could you also make sure examples compile in the github action? We just had a regression I didn't notice in examples. They don't need to run, just compile. Thanks!

Skyfold:
The `bin/dev` will work the same outside of nix, `nix develop` only brings in the dependencies. I updated it to use the `ghcid` command that runs the live reload server for examples (though you need to manuall refresh the browser).

I looked up `livereload` and see a few related projects, but am not sure which one you meant.

The `nix flake check` command builds the `examples` and `docgen` executables, so the CI will test them with both GHC versions.

The node2nix project for pulling in the nodejs dependencies needs `"lockfileVersion": 2`, so I left it out for now.

I needed to pull in `safe` for examples to compile with GHC 9.6.6 package set and I added `hSetBuffering stdout LineBuffering` so the `putStrLn` output does not get garbled.

This is good to merge. We can always add `livereload` later on.

Skyfold:
Fixed a minor bug I wasn't seeing locally because nix had cached the check.

seanhess:
Awesome, thanks!

PR #53, opened 2024-12-21
Add textarea element

tusharad:
Hi @seanhess,

This PR introduces a convenience function for the textarea element that triggers an action when the text changes. I've also added the necessary documentation and made the function accessible in Web.Hyperbole. Please let me know if there are any additional requirements or suggestions.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,3 +182,4 @@ Contributors
 * [Sean Hess](seanhess)
 * [Kamil Figiela](https://github.com/kfigiela)
 * [Christian Georgii](https://github.com/cgeorgii)
+* [Tushar Adhatrao](https://github.com/tusharad)
--- a/example/Example/Contact.hs
+++ b/example/Example/Contact.hs
@@ -66,6 +66,7 @@ data ContactForm f = ContactForm
   { firstName :: Field f Text
   , lastName :: Field f Text
   , age :: Field f Int
+  , info :: Field f Text
   }
   deriving (Generic)
 instance Form ContactForm Maybe
@@ -73,8 +74,8 @@ instance Form ContactForm Maybe
 
 parseUser :: (Hyperbole :> es) => Int -> Eff es User
 parseUser uid = do
-  ContactForm{firstName, lastName, age} <- formData @ContactForm
-  pure User{id = uid, isActive = True, firstName, lastName, age}
+  ContactForm{firstName, lastName, age, info} <- formData @ContactForm
+  pure User{id = uid, isActive = True, firstName, lastName, age, info}
 
 
 contactView :: User -> View Contact ()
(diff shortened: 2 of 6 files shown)


seanhess:
Hey @tusharad, welcome and thanks for contributing!

Can I ask about your UI? Is there a reason you didn't use a form instead? Forms are a little more complicated to use, but they are usually a better end-user experience for textareas than a live-updating field. 

What if we create a `textarea` that doesn't build in the onInput event: copying `input` (without the autocomplete stuff) from Web.Hyperbole.Forms. Textarea doesn't appear to use `value`, but the initial contents are a text node inside. Maybe add `Text` content as the last parameter?

How it would look to use it:

```
data MyForm f = MyForm
  { comments :: Field f Text
  } deriving (Generic)
instance Form MyForm Maybe

instance IOE :> es => HyperView MyView es where
  data Action MyView = Submit
  action Submit =
    myData <- formData @MyForm -- MyForm Identity
    print $ myData.comments
    pure _
    
myView :: View MyView ()
myView = do
  f <- formFields @MyForm
  form Submit id $ do
    field f.comments id $ do
      label "Comments"
      textarea id (placeholder "Comments here") "Initial Content"
```

With the new `onInput` event, you could still implement a live-updating text area, but it wouldn't be required. 

```
instance IOE :> es => HyperView MyView es where
  data Action MyView = LiveUpdate Text
  ...

myView View MyView ()
myVIew = do
  textarea (onInput LiveUpdate) "Initial Content"
```

I haven't had a lot of feedback on the new forms interface. I'm curious why your initial reaction was to avoid it. Too complicated?

tusharad:
Hi @seanhess, thank you for the feedback!

Initially, I found the Forms module a bit complicated and rigid, much like my experience with [Purescript Halogen Formless](https://github.com/thomashoneyman/purescript-halogen-formless). However, after diving deeper into the Forms module, I gained a better understanding and have since migrated my code to use [Forms](https://github.com/tusharad/Reddit-Clone-Haskell/blob/7714874407154686d412694ef26602b3d9f2664d/haskread-platform-ui/src/Platform/View/CommentCard.hs#L204). Adding examples that showcase textarea and checkbox usage in Forms could help others navigate the initial learning curve.

Regarding this PR, I believe the textarea function could work well as a standalone function, similar to dropdown. However, if it doesn’t align with the project’s design principles or priorities, feel free to close the PR. I appreciate the opportunity to contribute and am happy to discuss further.

Thanks again!

seanhess:
Great, I'm glad to hear you were able to figure it out. Forms are such a pain: you can either make them intuitive or easy to use, but not both. 

Would you be interested in updating this PR to implement `textarea` as a part of `Web.Hyperbole.Forms`? You would need to copy `input` (but remove the autocomplete stuff, and add default text content). 

I'd also love to see it used in one of the examples. I don't think it makes sense in the Signup Form example, but it could make sense to add notes for a contact in Example.Contact.hs.

tusharad:
Sure!

tusharad:
Hi @seanhess , I have made requested changes and added `textarea` example in contacts. Let me know if you need anything else. Thanks!

tusharad:
Done!

Issue #38, opened 2024-10-29
Feature Request // Keyboard OnKey listener Event


benjaminweb:
**Context:**
Building search input field with on-type updating auto-suggest list I can navigate with arrows like 

<img width="364" alt="Screenshot 2024-10-29 at 12 07 55" src="https://github.com/user-attachments/assets/f6744911-b1f2-4816-8533-b7d759bed988">

(from https://mullvad.net/en/servers)

**Requirements:**
Using arrows I want to be able to navigate through remaining suggestions.
That requires me to catch some specific key events, similarly to https://github.com/huytd/kanelm/blob/99b1d858e8abe18f54adb55a14f3c6cbf3629286/src/EventHelpers.elm#L9.

seanhess:
Is that an HTML `<select>` tag? I haven't made an autocomplete in a while, do plugin authors implement the key presses by hand or display a dropdown?

benjaminweb:
I don't see a dropdown (nor a HTML select tag) anywhere in the site's source:
<img width="662" alt="Screenshot 2024-10-29 at 23 29 08" src="https://github.com/user-attachments/assets/8b9a6277-3d38-4255-890b-75894751778c">

I haven't made one at all yet.

seanhess:
@benjaminweb I've implemented a half-hearted version of this: [Example/Search.hs ](https://github.com/seanhess/hyperbole/blob/main/example/Example/Search.hs)

It doesn't allow keyboard navigation yet. 

But maybe auto-complete is a good example of when one ought to use JS bindings. In #25 we talk a little about how to add easy interop with JS. It would be nice to be able to drop in a JS component like autocomplete and have it only fire Hyperbole actions when the user selects an item.

seanhess:
Added support in #56. Also added full autocomplete to [Example/Search.hs](./example/Example/Search.hs). Give it a spin and let me know what you think

PR #56, opened 2024-12-24
Autocomplete, OnKeyDown / OnKeyUp support

seanhess:
Created a full Autocomplete example. See examples/Example/Search.hs. View it in action here: https://docs.hyperbole.live/. Compare to Example/Filter.hs

Added support for OnKeyDown to support Enter key and up/down arrows


seanhess:
Pushed support for OnKeyUp as well. Fixes #38

Issue #32, opened 2024-10-17
Feature request: batteries-included way to encode record type to query args and back


benjaminweb:
it would be nice to have batteries included to generate query args from types and read them back in (currently show breaks that because it does not capitalise the data constructor's names).
Goes hand in hand with https://github.com/seanhess/hyperbole/issues/29

seanhess:
@benjaminweb - I just added [Web.Hyperbole.Effect.QueryData](https://github.com/seanhess/hyperbole/blob/main/src/Web/Hyperbole/Effect/QueryData.hs). Can you check to see if this solves your issue?

benjaminweb:
see #68

seanhess:
Take a look at #69, I'd love your feedback

PR #57, opened 2024-12-29
Added reqParamMaybe function

tusharad:
Hi @seanhess ,

This PR introduces a new function, `reqParamMaybe`, which is a Maybe version of the existing reqParam function. It is designed to provide flexibility by returning a Maybe type, enabling better handling of optional parameters in requests.

Please let me know if there are any changes or additional requirements needed.

Thanks!

### DIFF
--- a/src/Web/Hyperbole/Application.hs
+++ b/src/Web/Hyperbole/Application.hs
@@ -22,7 +22,7 @@ import Effectful
 import Effectful.Concurrent.Async
 import Effectful.Dispatch.Dynamic
 import Effectful.Error.Static
-import Network.HTTP.Types (parseQuery)
+import Network.HTTP.Types (parseQueryText)
 import Network.Wai qualified as Wai
 import Network.Wai.Handler.WebSockets (websocketsOr)
 import Network.WebSockets (Connection, PendingConnection, defaultConnectionOptions)
@@ -97,7 +97,7 @@ socketApp actions pend = do
     parse url cook hst mbody = do
       (u, q) <- parseUrl url
       let path = paths u
-          query = parseQuery (cs q)
+          query = parseQueryText (cs q)
           cookies = parseCookies $ cs $ header cook
           host = Host $ cs $ header hst
           method = "POST"
--- a/src/Web/Hyperbole/Effect/Event.hs
+++ b/src/Web/Hyperbole/Effect/Event.hs
@@ -1,12 +1,14 @@
 module Web.Hyperbole.Effect.Event where
 
+import Control.Monad (join)
+import Data.String.Conversions
 import Data.Text (Text)
 import Effectful
+import Network.HTTP.Types (QueryText)
 import Web.Hyperbole.Effect.Hyperbole (Hyperbole)
-import Web.Hyperbole.Effect.Request (lookupParam, reqParams)
+import Web.Hyperbole.Effect.Request (reqParams)
 import Web.Hyperbole.Effect.Server (Event (..))
 import Web.Hyperbole.HyperView (HyperView (..), ViewAction (..), ViewId (..))
-import Web.View (Query)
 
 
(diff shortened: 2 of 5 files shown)


seanhess:
Hmm, you're right that there's a hole here. We need to improve how all these functions work together. I like how the word "lookup" implies a `Maybe`.  Can we rename your function to have "lookup" int the name instead of "Maybe"?

It also looks like the type of the Query Parameter is inconsistent. We currently have:

```
reqParams :: (Hyperbole :> es) => Eff es Query
reqParam :: forall a es. (Hyperbole :> es, FromHttpApiData a) => Text -> Eff es a
lookupParam :: BS.ByteString -> Query -> Maybe Text
```

Let's refactor to make them all either `Text` or `ByteString`.  

Also, `lookupParam` feels fairly useless right now, since it doesn't run `parseQueryParam`. 


```
-- This could be Text, as long as it is consistent
type QueryParam = BS.ByteString

reqParams :: (Hyperbole :> es) => Eff es Query
reqParam :: (Hyperbole :> es, FromHttpApiData a) => QueryParam -> Eff es a

-- This is your function. What should it be named?
lookupParam :: (Hyperbole :> es, FromHttpApiData a) => QueryParam -> Eff es (Maybe a)

-- Do we want to get rid of this? Is there a good way to create a `parseParam` function instead that can be used by either of the above?
lookupParam :: QueryParam -> Query -> Maybe Text
```

Do you mind thinking about this and revising the PR? We can sneak it into the 0.4 release if it's done soon.

tusharad:
Hi @seanhess 

```haskell
type QueryParam = Text

-- | Return the entire list of 'QueryParam's
reqParams :: (Hyperbole :> es) => Eff es [(QueryParam, Text)]

-- | For a given 'QueryParam', return the value else throw an error
reqParam :: (Hyperbole :> es) => QueryParam -> Eff es Text

-- | For a given 'QueryParam', return the value else 'Nothing'
lookupParam :: (Hyperbole :> es) => QueryParam -> Eff es (Maybe Text)

-- | Whether the param is present or not
hasParam :: QueryParam -> Query -> Bool
```

What do you think of above names and types. Please do let me know any changes are required. I can go forward with that and push the commit.

seanhess:
I like the direciton. You might struggle converting the params to `[(Text, Text)]` preemptively. They start as `[ByteString, Maybe ByteString]`, reflecting that a parameter value can be empty (flags). 

Also, we definitely need functions that parse them for us. Like how `reqParam` automatically resolves `FromHttpApiData`. In your proposal how would that happen?

tusharad:
Instead of [(Text, Text)], we can use [(Text, Maybe Text)]. This aligns directly with how parameters are structured, accounting for cases where a parameter value may be empty (such as flags). Using the [queryToQueryText](https://hackage.haskell.org/package/http-types-0.12.4/docs/Network-HTTP-Types-URI.html#v:queryToQueryText) function , we can transform [ByteString, Maybe ByteString] into [(Text, Maybe Text)] efficiently. This also avoids the need for manual encodeUtf8 and decodeUtf8 calls, as queryToQueryText handles the conversion to Text for us. With this change, the reqParam function can continue to operate as it currently does

seanhess:
👍 I like it. 

Still need to figure out parameter parsing. It might be worth thinking about this issue along the way: https://github.com/seanhess/hyperbole/issues/32

tusharad:
Replaced [ByteString, Maybe ByteString] query representation with QueryText ([(Text, Maybe Text)]), using queryToQueryText for conversion. This simplifies handling of query parameters, ensuring consistent and native Text processing.

Functions like reqParams and lookupParam now directly operate on QueryText, eliminating the need for manual encodeUtf8 and decodeUtf8.

Let me know if further refinements are needed! Thanks!

seanhess:
Looks great! I might think harder and tweak naming later but nice work!

PR #62, opened 2025-01-04
Escape quotes to avoid broken link to "Components"

cgeorgii:
https://haskell-haddock.readthedocs.io/latest/markup.html#special-characters


seanhess:
Thanks!

PR #68, opened 2025-01-05
add query params example

benjaminweb:
Basic query params example; should I be able to read it directly to the record?
I've tried to use [formParse](https://github.com/seanhess/hyperbole/blob/f61bf9d420a95f9937623d5ce1815f9694c7c438/test/Test/FormSpec.hs#L35) to no avail.

The simplest user interface I can conceive is:
- data record (specifying which field is required and which is Maybe)
- specifying whether it can originate from POST or GET args
- command that decodes it from the request
(so, what I've in mind is a unified interface for both POST and GET)

### DIFF
--- a/example/Example/AppRoute.hs
+++ b/example/Example/AppRoute.hs
@@ -19,6 +19,7 @@ data AppRoute
   | Filter
   | Sessions
   | Requests
+  | QueryParams
   | Redirects
   | RedirectNow
   | LazyLoading
--- a/example/Example/Page/QueryParams.hs
+++ b/example/Example/Page/QueryParams.hs
@@ -1,0 +1,52 @@
+module Example.Page.QueryParams where
+
+import Data.Map qualified as Map
+import Data.Maybe (mapMaybe)
+import Data.String.Conversions (cs)
+import Data.Text (Text)
+import Data.Text qualified as T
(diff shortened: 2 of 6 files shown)


seanhess:
Nice work! I'm actually working on #32 right now, so the interface will change, but I'd love a good example for it. 

Ideally the example would be a page that demonstrated how to use query params to store state. At work, we have a page which shows all the items in our database, and allows the user to filter them based on various critiera with dropdowns and toggle buttons. 

An example might be something like Example.Page.Filter, but also has a toggle button to filter functional vs object oriented languages. It would store the search term and the other filter parameters in the query, allowing the user to reload the browser or share a link. 

I'll check something like the above in during my testing for #32, but feel free to expand on it or come up with a better example. I do think multi-faceted search might be the best use case.

seanhess:
Check out the new interface in #69. Any thoughts? I'd like to add an example as described above that leverages ToQuery / FromQuery

seanhess:
Sorry if I stepped on your toes, but I needed to add an example to test the new query params code. I updated Example.Page.Filter to add filters for the language family and type system features. These are all encoded to the query, and read to the query when loading

I'll close this PR for now, but feel free to reopen or open a new one if you come up with a more robust example

PR #69, opened 2025-01-05
Query Params - Serialization, update browser, sessions, etc

seanhess:
Implements #32 

Unified the interface to sessions and query parameters. They both rely on a common interface defined in Web.Hyperbole.Data.QueryData:

```
class ToParam a where
  toParam :: a -> Text

class FromParam a where
  parseParam :: Text -> Either Text a

-- from Web.Hyperbole.Effect.Query
param :: (Hyperbole :> es, FromParam a) => Text -> Eff es a
lookupParam :: (Hyperbole :> es, FromParam a) => Text -> Eff es (Maybe a)
setParam :: (Hyperbole :> es, ToParam a) => Text -> a -> Eff es ()
deleteParam :: (Hyperbole :> es) => Text -> Eff es ()

-- from Web.Hyperbole.Effect.Session
sessionKey :: (FromParam a, Hyperbole :> es) => Text -> Eff es a
lookupSessionKey :: (FromParam a, Hyperbole :> es) => Text -> Eff es (Maybe a)
deleteSessionKey :: (Hyperbole :> es) => Text -> Eff es ()
setSessionKey :: (ToParam a, Hyperbole :> es) => Text -> a -> Eff es ()
```

In order to save something to either the session or a param, it must implement `ToParam`, and `FromParam` to read it. 

There are also functions to save an entire object to either of them:

```
class ToQuery a where
  toQuery :: a -> QueryData

class FromQuery a where
  parseQuery :: QueryData -> Either Text a

query :: (Hyperbole :> es, FromQuery a) => Eff es a
setQuery :: (Hyperbole :> es, ToQuery a) => a -> Eff es ()

session :: (FromQuery a, Hyperbole :> es) => Eff es a
setSession :: (Hyperbole :> es, ToQuery a) => a -> Eff es ()
```

You can derive ToQuery / FromQuery using generics:

```
data MyFilters = MyFilters
  { term :: Maybe Text
  , showActiveOnly :: Maybe Bool
  } deriving (Generic, ToQuery, FromQuery)
```

Then you can load and save them in page an update

```
page :: Hyperbole :> es => Page es '[]
page = do
  filters <- query @MyFilters
  pure $ mainView filters
  
instance HyperView SomeView es where
  data Action SomeView = SetFilters MyFilters
  update (SE filters) = do
    setQuery filters
    pure $ mainView filters
```


benjaminweb:
- `ToParam, toParam` & `FromParam, parseParam`? Shouldn't it be `fromParam` instead of `parseParam`? Or `paramToQuery` and `queryToParam` to be more explicit?
- I took some seconds to figure that I build on top of `ToParam` `ToQuery` and `FromQuery`. Which alternative to `ToParam` would make it clearer that it's the foundation for `ToQuery` and `FromQuery`? Please correct me if I misunderstand it the way it is meant!
  - _What's the difference between `Query` and `Param`? What the translation between them?_
- `QueryData.insert "term" f.term` why not instead `f.term ..> "term"`
- similarly: `QueryData.require "term" q` why not instead `"term" <.. q`
- haven't tested the update browser location / url yet => how do I use it?
- no experience with sessions yet, therefore can't say what's the impact for me
- amended #68  to this PR (I didn't compile check as I cancelled getting your PR into my fork):

```haskell
{-# LANGUAGE RecordWildCards #-}

module Example.Page.QueryParams where

import Data.String.Conversions (cs)
import Data.Text (Text)
import Data.Text qualified as T
import Data.Text.Encoding (decodeUtf8)
import Effectful
import Example.AppRoute qualified as Route
import Example.View.Layout (exampleLayout)
import Web.Hyperbole
import Web.Hyperbole.Data.QueryData as QueryData
import Web.Hyperbole.Effect.Server (Host (..), Request (..))

data Dolphin = Dolphin
  { name :: Text,
    age :: Int
  }
  deriving (Eq, Show, ToParam)

instance ToQuery Dolphin where
  toQuery f =
    mempty
      & QueryData.insert "name" f.name
      & QueryData.insert "age" f.age

instance FromQuery Dolphin where
  parseQuery q = do
    name <- QueryData.require "name" q
    age <- QueryData.require "another" q
    pure Dolphin {..}

page :: (Hyperbole :> es) => Eff es (Page '[])
page = do
  r <- request
  pure $ exampleLayout Route.QueryParams $ do
    col (gap 10 . pad 10) $ do
      el_ $ do
        case readDolphin r.query of
          Right d -> do
            col (gap 10) $ do
              el_ $ text "successfully constructed record from query params:"
              el_ $ text $ cs $ show d
          Left _e -> col (gap 10) $ do
            el_ $ text $ cs ("failure: check query params" ++ show r.query)
            let Host host = r.host
            let queryArgs = T.intercalate "&" $ map (\(x, y) -> x <> "=" <> y) [("age", "12"), ("name", "bubu")]
            let ps = T.intercalate "/" $ "http://" : decodeUtf8 host : (r.path ++ ["?" <> queryArgs])
            let theUrl = url ps
            row (gap 10) $ do
              el_ $ text "click link to send query params:"
              el_ $ link theUrl id (text $ cs $ show theUrl)

```

seanhess:
> * `ToParam, toParam` & `FromParam, parseParam`? Shouldn't it be `fromParam` instead of `parseParam`? Or `paramToQuery` and `queryToParam` to be more explicit?

See Data.Aeson. FromJSON has a `parseJSON` function instead of a `fromJSON` function. Can you find examples of parsers that name it the way you suggest?



> * What's the difference between `Query` and `Param`? What the translation between them?

Well, the names come from the fact that for query params, each `Param` is encoded as a single parameter: message=hello. Whereas the Query is the whole query string with multiple parameters: message=hello&age=1. Sessions are encoded the same way. 





> * `QueryData.insert "term" f.term` why not instead `f.term ..> "term"`
> * similarly: `QueryData.require "term" q` why not instead `"term" <.. q`

I'm not a fan of introducing new operators for this. They don't buy us that much. The current interface mirrors `Data.Map`



> * haven't tested the update browser location / url yet => how do I use it?

It happens automatically if you call `setParam` or `setQuery`



> * amended [add query params example #68](https://github.com/seanhess/hyperbole/pull/68)  to this PR (I didn't compile check as I cancelled getting your PR into my fork):

You should be able to simplify your example quite a bit. Does your page only have one dolphin?

```

data Dolphin = Dolphin
  { name :: Text,
    age :: Int
  }
  -- changed to ToQuery instead of ToParam. Automatically derived for Text and Int fields. Will serialize to `name=xxx&age=xxx`
  deriving (Eq, Show, Generic, ToQuery)

page :: (Hyperbole :> es) => Eff es (Page '[])
page = do
  dolphin <- query
  pure $ exampleLayout Route.QueryParams $ do
    col (gap 10 . pad 10) $ do
        col (gap 10) $ do
           el_ $ text "successfully constructed record from query params:"
           el_ $ text $ cs $ show d
```

That won't handle missing parameters gracefully though, as the name and age are required.  I'll have to think about how to make that easier.

benjaminweb:
> > * `ToParam, toParam` & `FromParam, parseParam`? Shouldn't it be `fromParam` instead of `parseParam`? Or `paramToQuery` and `queryToParam` to be more explicit?
> 
> See Data.Aeson. FromJSON has a `parseJSON` function instead of a `fromJSON` function. Can you find examples of parsers that name it the way you suggest? I don’t. — I never liked `Data.Aeson`’s choice `parseJSON`. I understand your intention of aligning with established standards now; though, I am hesitant to always align to the most used variant — popularity is not always a sign of being best (end of rant).

> > * What's the difference between `Query` and `Param`? What the translation between them?
> 
> Well, the names come from the fact that for query params, each `Param` is encoded as a single parameter: message=hello. Whereas the Query is the whole query string with multiple parameters: message=hello&age=1. Sessions are encoded the same way.

Now I understand. How could we change the wording of `Param` and `Query` to reflect their very difference better?

> > * `QueryData.insert "term" f.term` why not instead `f.term ..> "term"`
> > * similarly: `QueryData.require "term" q` why not instead `"term" <.. q`
> 
> I'm not a fan of introducing new operators for this. They don't buy us that much. The current interface mirrors `Data.Map`

Fine for me!

> > * haven't tested the update browser location / url yet => how do I use it?
> 
> It happens automatically if you call `setParam` or `setQuery`

I would never have associated it from their names that they’d do that :). Only rescue would be their docstring or an example.

> > * amended [add query params example #68](https://github.com/seanhess/hyperbole/pull/68)  to this PR (I didn't compile check as I cancelled getting your PR into my fork):
> 
> You should be able to simplify your example quite a bit.

Your edited version is as straightforward as I desire it to be.

> Does your page only have one dolphin?

Which way you have in mind to add more?

> ```
> 
> data Dolphin = Dolphin
>   { name :: Text,
>     age :: Int
>   }
>   -- changed to ToQuery instead of ToParam. Automatically derived for Text and Int fields. Will serialize to `name=xxx&age=xxx`
>   deriving (Eq, Show, Generic, ToQuery)
> 
> page :: (Hyperbole :> es) => Eff es (Page '[])
> page = do
>   dolphin <- query
>   pure $ exampleLayout Route.QueryParams $ do
>     col (gap 10 . pad 10) $ do
>         col (gap 10) $ do
>            el_ $ text "successfully constructed record from query params:"
>            el_ $ text $ cs $ show d
> ```
> 
> That won't handle missing parameters gracefully though, as the name and age are required. I'll have to think about how to make that easier.

seanhess:
> Now I understand. How could we change the wording of `Param` and `Query` to reflect their very difference better?
> What if we name them `QueryParamsFullstring` and `QueryParamsPairs` (or a similar way)?

These already reflect the naming in the HTTP standard. From https://en.wikipedia.org/wiki/Query_string
"A query string is a part of a uniform resource locator (URL) that assigns values to specified parameters"

I could be convinced to rename them to something like `ToParamValue` and `ToQueryString`. Would that be more clear?

seanhess:
> > > * haven't tested the update browser location / url yet => how do I use it?
> > 
> > 
> > It happens automatically if you call `setParam` or `setQuery`
> 
> I would never have associated it from their names that they’d do that :). Only rescue would be their docstring or an example.

What did you expect setParam and setQuery to do, if not set the browser url?

benjaminweb:
> > Now I understand. How could we change the wording of `Param` and `Query` to reflect their very difference better?
> > What if we name them `QueryParamsFullstring` and `QueryParamsPairs` (or a similar way)?
> 
> These already reflect the naming in the HTTP standard. From https://en.wikipedia.org/wiki/Query_string "A query string is a part of a uniform resource locator (URL) that assigns values to specified parameters"
> 
> I could be convinced to rename them to something like `ToParamValue` and `ToQueryString`. Would that be more clear?

`ToQueryString`, yes. `ToParamValue` not so. `KeyValue` would be natural to me. Do we need to stick to the word Param that implies key, value?

> What did you expect setParam and setQuery to do, if not set the browser url?
`setParam` would set a Param like `setQuery` set a Query. `setBrowserUrl` would set the browser's URL for me.

seanhess:
> `ToQueryString`, yes. `ToParamValue` not so. `KeyValue` would be natural to me. Do we need to stick to the word Param that implies key, value?

Param only implies it is being used as a value for a parameter. KeyValue would be confusing because it doesn't specify the key. I think I like the current naming better than any of the alternatives that have been proposed. 



> `setParam` would set a Param like `setQuery` set a Query. `setBrowserUrl` would set the browser's URL for me.

It does. `setQuery` also updates the browser url. What did you expect `setQuery` to do if not set the browser's query string? 

Thanks for all your feedback!

benjaminweb:
Thanks for bringing all things together and crafting it!

seanhess:
Merging!

Issue #76, opened 2025-01-21
How to update web-view?


seanhess:
Hey @Skyfold, the examples package.yaml requires web-view >= 0.7 now, but now the CI action isn't completing. How do I update it to get the latest from hackage? The framework itself works wth 0.6.2 or greater. I'm fine updating the requirement to 0.7 if that helps:

```
      > Using Parsec parser
       > Configuring hyperbole-examples-0.4.2...
       > CallStack (from HasCallStack):
       >   withMetadata, called at libraries/Cabal/Cabal/src/Distribution/Simple/Utils.hs:368:14 in Cabal-3.10.3.0:Distribution.Simple.Utils
       > Error: Setup: Encountered missing or private dependencies:
       > web-view >=0.7 && <1
       >
       For full logs, run 'nix log /nix/store/mv061w6nv5rp7p3sz1gcbjcccfcdxkld-hyperbole-examples-0.4.2.drv'.
```

seanhess:
Use branch https://github.com/seanhess/hyperbole/tree/web-view-0.7 to reproduce

Skyfold:
To update the dependencies that nix provides we need to run `nix flake lock --update-input <name_of_input>` or `nix flake update` to update all inputs. Both commands update the `flake.lock` file with the new hash and rev for the latest commit on the default branch. We can change the flake.nix to track a specific tag, commit or branch.

I see that "cc04bfe341d4a36d54e6d7463b63b71a6f0b44d6" is the commit for 0.7.0 for web-view. Since that commit is the latest commit, I went ahead and ran `nix flake update` and created a pull request #77

Skyfold:
The way I have written the flake.nix makes it harder to pull web-view from hackage. In the hyperbole flake.nix we import web-view's overlay which not only provides the web-view source, but also it's dependency overrides. So, if we switch the web-view source from git to hackage, I need to add the web-view's flake.nix overrides to hyperbole's flake.nix before hyperbole will compile.

I think having separate release branches for only major versions would make updating the web-view source for hyperbole easy. For minor changes: `nix flake lock --update-input web-view`. For major changes: change the branch for the web-view input in `flake.nix` and run `nix flake lock --update-input web-view`.

seanhess:
Thanks for getting to this. I was planning on using "main" as a dev branch, and "0.4" would be a release branch. Would that make it easier?

seanhess:
So I just pushed

Skyfold:
[...]

Issue #75, opened 2025-01-20
docs.hyperbole.live not automatically updating


seanhess:
Hey @Skyfold, I just pushed a couple of commits to main. The workflows are completing successfully, but the examples site isn't updating. The activity logs on digitalocean don't show anything happening

Skyfold:
Looks like we will need to push to Digital Ocean's registry to get auto deploy:

"App Platform does not support auto-deployment of images from Docker Hub repositories or from GitHub Registry"

I manually updated the app on digital ocean.

To push to Digital Ocean's registry we just need create a registry on the account and add its credentials to secrets. Then I'll update the workflow to push to both registries.

Skyfold:
I created a registry on Digital Ocean called hyperbole on the account since it is free for under 500mb. You just need to create a personal access token scoped to the registry and add its credentials to GitHub secrets. Name the secret DIGITAL_OCEAN_API.

seanhess:
Ok, I created a digital ocean api token with the registry scope, and added it here as a respository secret named DIGITAL_OCEAN_API. (There was also an option to add an "Environment" secret. Not sure what the difference is)

Skyfold:
I did some reading and I think we should be using Environment secrets. It provides a more fine grained way to confine secret access. The `DIGITAL_OCEAN_API` should only be available to commits in main. Right now it is available to all branches (just not pull requests).

We only want to update the docker image on docs.hyperbole.live on push to main. 

Can you create one environment: `prod`, set "Selected branches and tags" to main and add both DIGITAL_OCEAN_USERNAME and DIGITAL_OCEAN_API

The DIGITAL_OCEAN_USERNAME needs to be your digital ocean registered email.

I set the `docker` job to run with the `prod` environment.

The updated workflow is in pull request #78

seanhess:
Ok, I think I set this all up, can you verify?

DIGITAL_OCEAN_API is set in a prod environment secret. 
DIGITAL_OCEAN_USERNAME is set up in "prod" as both a secret and a variable. Set to my digital ocean email

Skyfold:
Go ahead and re-run the last action in the Haskell CI. If we see that the "Last Deployed" on digital ocean's App Platform updates, then it worked.

We should also see the docker job running "Log into Digital Ocean Registry" and "Push to Digital Ocean".

seanhess:
[...]

PR #84, opened 2025-01-31
Fixes DO docker registry push and uses web-view from hackage 0.7.0

Skyfold:
This should push to the digital ocean docker registry, if the credentials are there. I got the workflow to run, but I kept getting timeouts when pushing. I'm wondering if it needs to come from your account instead of mine.

I also pull the source for web-view from hackage:
  - Will need to wait until the next release before tracking 0.7 with the hyperbole flake since the commits that update web-view's flake are in-between the commit that adds FlexWrap. 
  - `nix flake update web-view` won't work in the meantime.
  - Uses web-view 0.7.0
  - Does not run the web-view tests since the source on Hackage is missing the test files in resources.

Adresses #76 #75


Skyfold:
I think we should switch back to using docker for CI, see comments in #76.

PR #91, opened 2025-03-09
Fixed source links in example app

AphonicChaos:
At the moment, clicking any "View Source" links in the example app will result in a 404 because for the 0.4 tag, the directory is `example` not `examples`. The other alternative would be to update the 0.4 tag to point to the latest version of the repo, which does indeed have an `examples` directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Revision history for hyperbole
 
+## 0.4.4 -- 2025-03-09
+
+* fixed source links in examples
+
 ## 0.4.3 -- 2024-01-31
 
 * Bug fixes and improvements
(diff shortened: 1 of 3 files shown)


seanhess:
Hey, thanks for making this a PR! I need to switch the example site to use the release branch instead of main, but that can wait until after we merge this and get it working again.

seanhess:
I'm very slowly becoming more disciplined with releases. I started with a "latest" branch, but switched to naming release branches by their major version, like 0.4. So latest is stale. I'll delete it. 

If you have a better idea of how to improve the process I'm all ears! We currently have a chaotic main branch for development, and want to avoid breaking the examples site, including source code links.

Issue #97, opened 2025-03-22
TodoMvc example is incomplete and questions arise


benjamin-thomas:
Hello!

Hyperbole looks quite interesting!

I've been playing with the framework for a few days, and I have a few questions.

I can spot a couple of problems in the todomvc example:

1. the user input is not cleared after submitting a new todo
2. the "items left" message does not update after checking/unchecking a todo

Regarding point number 1, it looks like the framework does not have a good way to deal with that scenario yet, is that correct?
Regarding point number 2, the problem seems to arise from the use of nested hyperviews (one per item): is it possible to "signal to the parent" hyperview that something happened which requires a state change on the parent side?

As a side note, I think it'd be nice to make the web-view code opt-in. Personally, I prefer outputting straight HTML and deal with the styling in CSS as much as possible. But in any case, it'd be nice to have the option (to opt-in or opt-out), to enable porting existing pages that already have a design, with less friction. Currently, it feels like hyperbole relies heavily on web-view which makes avoiding it a little cumbersome.

Here's my attempt at implementing the todomvc example, which triggers a couple more questions:

```hs
{-# LANGUAGE DerivingStrategies #-}
{-# LANGUAGE QuasiQuotes #-}
{-# LANGUAGE UndecidableInstances #-}

module Main where

import Control.Monad (forM_)
import Data.ByteString.Lazy qualified as BSL
import Data.Map (Map)
import Data.Map.Strict qualified as M
import Data.Maybe
import Data.String.Interpolate
import Data.Text (Text)
import Data.Text qualified as T
import Effectful
import Effectful.Concurrent.STM
import Effectful.Dispatch.Dynamic
import Web.Hyperbole hiding (basicDocument)
import Web.Hyperbole.Data.QueryData qualified as QD
import Web.Hyperbole.HyperView (toAction)
import Web.View.Style
import Web.View.Types (AttValue)
import Prelude hiding (div, span)

{-
rg --files -t hs -g '!/dist*' | entr -rc cabal run todo2
ghcid -c 'cabal repl todo3' -T :main
-}

data Todo = MkTodo
    { id_ :: Int
    , descr :: Text
    , done :: Bool
    }
    deriving (Show, Read, Eq)

main :: IO ()
main = do
    let port = 4321 :: Int
    putStrLn $ "Starting web server on port: " <> show port
    run port app

updateVarReturning ::
    (Concurrent :> es) =>
    TVar a ->
    (a -> a) ->
    Eff es a
[...]

PR #98, opened 2025-03-22
Fixed: Example TodoMVC - items left count doesn't update on toggle click

seanhess:
As mentioned in #97, the item count wasn't updating in the TodoMVC example when the user clicked the toggle button. 

* Moved `SetCompleted` to the parent `AllTodos` view, and targeted it with the button. 

This might be our motivating example for #15! Passing FilterTodo around was annoying, and this would be cleaner with `trigger`

### DIFF
--- a/examples/Example/Page/Todo.hs
+++ b/examples/Example/Page/Todo.hs
@@ -53,6 +53,7 @@ instance (Todos :> es) => HyperView AllTodos es where
     | Filter FilterTodo
     | SubmitTodo
     | ToggleAll FilterTodo
+    | SetCompleted FilterTodo Todo Bool
     deriving (Show, Read, ViewAction)
 
   update = \case
@@ -71,6 +72,10 @@ instance (Todos :> es) => HyperView AllTodos es where
     Filter filt -> do
       todos <- filteredTodos filt
       pure $ todosView filt todos
+    SetCompleted filt todo completed -> do
+      _ <- Todos.setCompleted completed todo
+      todos <- filteredTodos filt
+      pure $ todosView filt todos
    where
     filteredTodos filt =
       filter (isFilter filt) <$> Todos.loadAll
@@ -86,7 +91,7 @@ todosView filt todos = do
   todoForm filt
   col id $ do
     forM_ todos $ \todo -> do
-      hyper (TodoView todo.id) $ todoView todo
+      hyper (TodoView todo.id) $ todoView filt todo
   statusBar filt todos
 
 todoForm :: FilterTodo -> View AllTodos ()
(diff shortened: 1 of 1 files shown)


examples/Example/Page/Todo.hs
```diff
@@ -133,34 +138,31 @@ instance (Todos :> es) => HyperView TodoView es where
...
+    pure $ todoView filt updated
 
-todoView :: Todo -> View TodoView ()
-todoView todo = do
+todoView :: FilterTodo -> Todo -> View TodoView ()
+todoView filt todo = do
   row (border (TRBL 0 0 1 0) . pad 10) $ do
-    toggleCheckBtn (SetCompleted todo) todo.completed
-    el (completed . pad (XY 18 4) . onDblClick (Edit todo)) $ text todo.task
+    target AllTodos $ do
```
benjamin-thomas, line 155: This target mechanism is great, I like it!

examples/Example/Page/Todo.hs
```diff
@@ -133,34 +138,31 @@ instance (Todos :> es) => HyperView TodoView es where
...
+      toggleCheckBtn (SetCompleted filt todo) todo.completed
+    el (completed . pad (XY 18 4) . onDblClick (Edit filt todo)) $ text todo.task
  where
   completed = if todo.completed then Style.strikethrough else id
 
-todoEditView :: Todo -> View TodoView ()
-todoEditView todo = do
+todoEditView :: FilterTodo -> Todo -> View TodoView ()
```
benjamin-thomas, line 161: Just food for thought, but since you mentioned that passing around the filter is annoying, I'm wondering if it would be possible to express a "model" (elm-style) that would typically go with a hyperview instance and be automatically serialized/deserialized in the update function (backed by a cookie store I suppose?)

Maybe I missed it, but it doesn't look like we can rely on data to be present in the browser's. So maybe that approach would make the "trigger" proposition not needed (which I don't fully grasp yet)

benjamin-thomas (commented):
Just a little feedback

examples/Example/Page/Todo.hs
seanhess: Avoiding a model was a difficult design decision. Ultimately it wasn’t worth the cost. It’s less annoying to pass small bits of state around than to build in a model. 

The design philosophy of Hyperbole is closer to HTMX than it is to Elm. I think it’s a good trade off, both making the framework more intuitive, and I think it’s easier to work with for experienced users too

seanhess: > Maybe I missed it, but it doesn't look like we can rely on data to be present in the browser's.

I don’t understand, what do you mean by data being present in the browser? Hyperbole has built in support for state persistence via cookies/sessions, query parameters, and passing state in ViewActions. Is that what you’re talking about?

seanhess: Oh I see you’re talking about persisting a Model. An old version of this years ago kept track by serializing it, then passing it back to the server on a request. It worked great, but I decided to go with a simpler Action -> Updated View for this. Instead of Action -> Reduce Model -> Updated View. 

It’s simpler, easier to work with so far (I worked in Elm, React, and others for years)

benjamin-thomas: Yes, I was talking about persisting the Model. Sounds good to me though, I'm more familiar with Elm than anything, so I have to readjust my mental model a bit :+1:

PR #99, opened 2025-03-22
Manually set value attribute, autofocus

seanhess:
Encouraged by @benjamin-thomas, I updated the client to force inputs to their value attribute after update. This allows you to clear inputs and more. 

* Manually set input.value to whatever the attribute value is
* Filter and Autocomplete Example - added a clear button to the search field
* TodoMVC, Contacts Example - clear form after submission
* Manually focus the first element with [autofocus]
* Manually set input.checked to the value of dataset.checked. Added `checked` helper. 

TODO:
- [x] Form Parsing Checkboxes is broken!
- [x] Is input[type=checkbox] correct for a generic toggle? (REMOVED)


seanhess:
Added support for checkboxes, a toggle as a checkbox, etc.

Issue #100, opened 2025-03-24
Support External CSS (Example)


seanhess:
In #97 @benjamin-thomas tried to implement TodoMVC using their external stylesheet and found it confusing. 

Create an example implementing TodoMVC using their stylesheet. 

Refactor as necessary to make it not terrible

benjamin-thomas:
Hi Sean!

I've been playing with your library over the week-end and I'd be happy to contribute an example if you think it'd be useful.

My working example is here: https://github.com/benjamin-thomas/7guis/tree/063a04d7c3bf00fe230b245a124454a3853d7fbe/haskell/hyperbole/app

> Can you think of a better way to handle imports? 

Regarding this question (from https://github.com/seanhess/hyperbole/issues/97#issuecomment-2748413578), I'm thinking it'd be nice to have an alternative "global import", something like `import Web.Hyperbole.Minimal`, that would only put the bare essentials into scope. The golden path would be to `import Web.Hyperbole`, not `import Web.Hyperbole.Minimal`. How does that sound?

Also, if you're open to examples contributions, it'd be interested in contributing ["7 GUI"](https://eugenkiss.github.io/7guis/) examples, which I think is an interesting benchmark because it's simple enough not to be a time sink, but complex enough to demonstrate tricky interactions and workflows.

Although I've never used vuejs, I've noticed that their doc references the 7 GUI examples (and the traditional todomvc), here: https://vuejs.org/examples/#counter

So, it could be a good way to demonstrate what the library can do, and also to demonstrate useful patterns.

Either separately from the docs.hyperbole.live site, or integrated into it?

I'd expect some examples to require a little bit of javascript to work, which would be interest to show as well.

seanhess:
It would be great to have your help!

> My working example is here: https://github.com/benjamin-thomas/7guis/tree/063a04d7c3bf00fe230b245a124454a3853d7fbe/haskell/hyperbole/app

If you'd like to submit a PR for the TodoMVC external CSS example, can you do one that only changes the views, but reuses the existing Effect and HyperViews? I believe only the view functions need to change. 



> Regarding this question (from [#97 (comment)](https://github.com/seanhess/hyperbole/issues/97#issuecomment-2748413578)), I'm thinking it'd be nice to have an alternative "global import", something like `import Web.Hyperbole.Minimal`, that would only put the bare essentials into scope. The golden path would be to `import Web.Hyperbole`, not `import Web.Hyperbole.Minimal`. How does that sound?

[...]

PR #102, opened 2025-03-25
WIP: Add external css example

benjamin-thomas:
Attempt to demonstrate how one could use Hyperbole, with just CSS, see #100 

The idea is to demonstrate side-by-side the same implementation of the todomvc example.

For the moment, I've managed to inject a custom head via path detection before the library code starts.

I've then extracted common logic to a "shared" module, but I'm not sure how to go about sharing the hyperviews, since they depend on the views themselves, which need to be unique to the 2 module implementations.

Currently, I've renamed some types, e.g. `AllTodos` => `MkTodosView` to facilitate my own understanding, but I'll revert this at the end.


examples/Example/App.hs
```diff
@@ -75,19 +80,57 @@ import Web.Hyperbole.Effect.Server (Request (..))
 run :: IO ()
 run = do
   hSetBuffering stdout LineBuffering
-  putStrLn "Starting Examples on http://localhost:3000"
+
+  port <- do
```
seanhess, line 84: Please put any unreleated improvements / refactors into a separate PR

examples/Example/App.hs
```diff
@@ -75,19 +80,57 @@ import Web.Hyperbole.Effect.Server (Request (..))
...
-        app users count
+      Static.staticPolicy (addBase "examples/static") $ \req -> do
+        let doc = case pathInfo req of
+              "todoscss" : _ -> toDocumentTodosCSS
+              _ -> toDocument
+        (app doc users count) req
+
+-- Use the embedded version for real applications (see basicDocument).
+-- The link to /hyperbole.js here is just to make local development easier
+toDocument :: BL.ByteString -> BL.ByteString
```
seanhess, line 102: No need to branch here. You could either load your custom css on every page by modifying `toDocument`, or, even better, do everything in TodoCSS.page:

```
page :: (Todos :> es) => Eff es (Page '[TodosView, TodoView])
page = do
  todos <- Todos.loadAll
  pure $ tag "div" id $ do
     stylesheet "https://cdn.jsdelivr.net/npm/todomvc-app-css@2.4.3/index.min.css"
     -- the rest of your page
     ...
```

That way, your stylesheet is loaded on your page, and nothing else. All without modifying App.js

examples/Example/Page/Todos/Todo.hs
```diff
@@ -26,23 +27,17 @@ page = do
...
-simplePage = do
-  todos <- Todos.loadAll
-  pure $ do
-    hyper AllTodos $ todosView FilterAll todos
-
---- AllTodos ----------------------------------------------------------------------------
+--- TodosView ----------------------------------------------------------------------------
 
-data AllTodos = AllTodos
+data TodosView = MkTodosView
```
seanhess, line 34: Careful not to rename / refactor things to your preference in this PR. It'll make it hard to separate merging the example from a discussion on preferred naming.

examples/Example/Page/Todos/TodoCSS.hs
```diff
@@ -0,0 +1,147 @@
...
+import Example.Page.Todos.Shared
+import Example.Style qualified as Style
+import Example.View.Icon qualified as Icon
+import Example.View.Inputs (toggleCheckbox)
+import Web.Hyperbole as Hyperbole
+
+page :: (Todos :> es) => Eff es (Page '[TodosView, TodoView])
+page = do
+  todos <- Todos.loadAll
+  pure $ tag "div" id $ do
```
seanhess, line 22: See the comment in App.hs, but you can load your stylesheet here:

```
pure $ tag "div" id $ do
  stylesheet "https://cdn.jsdelivr.net/npm/todomvc-app-css@2.4.3/index.min.css"
  tag "h1" id $ text "Todos CSS"
  ...
```

examples/Example/Page/Todos/TodoCSS.hs
```diff
@@ -0,0 +1,147 @@
...
+    tag "h1" id $ text "Todos CSS"
+    tag "p" id $
```
seanhess, line 24: Why use `tag` repeatedly? Define aliases for any tags you are missing:

```
p :: Mod id -> View id () -> View id ()
p = tag "p"
```

examples/Example/Page/Todos/TodoCSS.hs
```diff
@@ -0,0 +1,147 @@
...
+  col id $ do
+    forM_ todos $ \todo -> do
+      hyper (MkTodoView todo.id) $ todoView filt todo
+  statusBar filt todos
+
+todoForm :: FilterTodo -> View TodosView ()
+todoForm filt = do
+  let f :: TodoForm FieldName = fieldNames
+  tag "div" id $ do
+    tag "span" id $ do
```
seanhess, line 86: Is "span" necessary here? (Does the CSS specifically target span?) Why not just use `el`? 

Perhaps you are also expressing a preference to control the exact html tags used / semantic DOM?

seanhess (changes requested):

seanhess:
great start! I added some comments and questions. 

I don't see any use of the external stylesheet yet?

I'm also not sure the best way to include both examples without code duplication. I'd really prefer not to duplicate any non-external-stylesheet code if possible, so the examples don't get out of sync. I'd also prefer to avoid making the examples too complicated or generalized. 

Since our TodoMVC with external CSS example is an extension of the normal one, maybe it will help it have it import the original example? Then at least you don't need the .Shared module, you can keep them in the original example and import from there.

examples/Example/App.hs
benjamin-thomas: I branched here because the `toDocument`  function of the example app injects CSS via the head tag, and that would interfere with the CSS loaded after.

image

examples/Example/Page/Todos/Todo.hs
benjamin-thomas: Yes, I said in the description

> Currently, I've renamed some types, e.g. AllTodos => MkTodosView to facilitate my own understanding, but I'll revert this at the end.

It looks like you didn't see it

examples/Example/Page/Todos/TodoCSS.hs
benjamin-thomas: Yes I will definitely create nicely named functions.

By opening the PR in draft mode, my intention was to get early feedback regarding the way I managed to inject the CSS rules for this specific page, and also maybe to get advice and sharing hyperview code wich looks tricky (but you already answered that, so I'll just duplicate code for now)

So the UI is not done currently, I've only implemented the bare essentials to move forward.

examples/Example/App.hs
seanhess: Hmm.... The issue is that Example.App.hs is serving double-duty as an example of how to bootstrap an application. I'd really like to avoid anything as complex as a branch like this if possible. 

This CSS is just a reset. Are you sure it would conflict? If it will definitely mess it up, maybe let's put this in its own app. I'm open to better ideas

examples/Example/Page/Todos/Todo.hs
seanhess: Got it thanks!

examples/Example/App.hs
benjamin-thomas: I went with that assumption but I didn't actually try it. I'll keep you posted.

benjamin-thomas:
I'm happy with the result currently @seanhess 

Maybe you can take a look to see if there are any snags left? And if not maybe as a last step I can try to fuse both examples into the same module and remove the shared module as discussed? Or maybe it's good enough with that bit of duplication?

I suspect the module itself may be a bit messy after that, so I reckon it's a good point to take a look now. Then I can try that last change or revert if it turns out not to be better.

examples/Example/Page/Todos/Todo.hs
```diff
@@ -28,13 +29,7 @@ page = do
...
       hyper AllTodos $ todosView FilterAll todos
 
-simplePage :: (Todos :> es) => Eff es (Page '[AllTodos, TodoView])
-simplePage = do
-  todos <- Todos.loadAll
-  pure $ do
-    hyper AllTodos $ todosView FilterAll todos
-
---- AllTodos ----------------------------------------------------------------------------
+--- TodosView ----------------------------------------------------------------------------
```
benjamin-thomas, line 32: This function seemed old/unused so I deleted it.

examples/Example/Page/Todos/TodoCSS.hs
```diff
@@ -0,0 +1,274 @@
...
+      input' -- we use a custom input field, because the Hyperbole one overrides autocomplete
+        ( extClass "new-todo"
+            {-
+              -- . autofocus
+              FIXME: turning off autofocus, that "steals" the focus on item click.
+              FIXME: to solve this, we could either store the "initially focused" state and track that boolean, or use buttons
+              FIXME: but since this example is meant to match as close as possible to the original CSS version
+              FIXME: and not diverge too much from the other todo example, I'm leaving as-is.
+             -}
+            . att "autocomplete" "off"
```
seanhess, line 138: We should go fix the source for `input`, right? Probably set to "off" most of the time. (Right now `input TextInput`, which is in the main example, sets it to "text-input" which is wrong). 

I wanted to make it easy to set "type" and "autocomplete" with one type, but maybe it wasn't a great choice?

examples/Example/Page/Todos/TodoCSS.hs
```diff
@@ -0,0 +1,274 @@
...
+    ul' (extClass "filters") $ do
+      filterLi FilterAll "All"
+      filterLi Active "Active"
+      filterLi Completed "Completed"
+    space
+    button ClearCompleted (extClass "clear-completed") "Clear completed"
+ where
+  filterLi f str =
+    li' (extClass "filter" . selectedFilter f) $ do
+      a
```
seanhess, line 167: Did you avoid using `Web.View.link` deliberately? Or just not know about it? `Web.View.Types.Url` has an `IsString` instance:

```
link "" (onClick (Filter f)) (text str)
```

seanhess:
Cool, thanks for doing this! I'll can play with it from here (I need to think about how to present this in a clear way that shows the minimum necessary changes between the two examples). 

The Big Question: how did it go? Were you fighting the framework too much? Which helpers or tools would make your experience smoother?

benjamin-thomas:
Great.

> The Big Question: how did it go? Were you fighting the framework too much? Which helpers or tools would make your experience smoother?

It went great! I definitely want to play with the library some more. As a learner of the language still, I've been on the lookout for a good Haskell web library for a while, and Hyperbole feels nice because it brings something to the table that's quite different from more traditional solutions. I've played a fair bit with Elixir/Phoenix prior to learning Haskell, but didn't feel compelled to invest into that technology because I was more interested in having a solid language rather than a cool tech. But I'm quite keen to explore a server-first approach to reactivity, it feels quite effective.

The only thing were I felt friction was the global import, that kinda made my understanding a bit muddy at the start.

What I'd like to understand next is: 

- what are the good patterns that play well with Hyperbole?
- what is the limit of this server-managed workflow, how far can I push it?
  - when/if I need more, how can I delegate to front-end logic gracefully?

seanhess:
Hey @benjamin-thomas, sorry to pull the rug out from under you, but I tagged you in a big refactor PR for web-view (now atomic-css). It'll directly affect your example. https://github.com/seanhess/atomic-css/pull/21

We're getting rid of the global import, but I'm not 100% sure what to replace it with. The "default" utility-based approach will be to import both hyperbole and atomic-css:

```
import Web.Atomic.CSS
import Web.Hyperbole
```

If you import only Hyperbole, you'll have access to the `@` operator and a `class_` attribute modifier that will merge classes for you, but that's it. We aren't exporting a full set of html tags or anything. 

LMK if you have any questions, but after reviewing the PR can you update this?

benjamin-thomas:
Hi @seanhess,

Yeah no problem! I intend to look at the refactor PR this week-end.

Then I'll bring this one up to date once it's ready.

I'll keep you posted

benjamin-thomas:
I've rebased my branch onto main and my original example mostly works now.

I still have a bit of work to do though so I'm putting the PR in draft for now (mainly cleanup and possible remove some unecessary duplication)

benjamin-thomas:
@seanhess I think it's ready if you want to take a look

- I added a missing "delete" action to both implementations
- I tried to centralize logic between both implementations via wrapper types.
- I've made it so the list is shared between both implementations

To counteract the CSS reset, I had already tweaked things with a few "style" attributes here and there. It's only minor though, so I don't think it's too much of a problem to leave it as such.

examples/Example/Page/Todos/Shared.hs
```diff
@@ -0,0 +1,88 @@
...
+data TodosAction
+  = ClearCompleted
+  | Filter FilterTodo
+  | SubmitTodo
+  | ToggleAll FilterTodo
+  | SetCompleted FilterTodo Todo Bool
+  | Destroy FilterTodo Todo
+  deriving (Generic, ViewAction)
+
+updateTodos :: (Todos :> es, Hyperbole :> es) => TodosAction -> Eff es [Todo]
```
seanhess, line 51: Nice factoring here!

benjamin-thomas: Thanks!

PR #116, opened 2025-08-07
OAuth2 Helper Effect

seanhess:
@adithyaov I decided hoauth2 wasn't doing enough to be worth the extra dependencies. I added `Web.Hyperbole.Effects.OAuth2` which integrates less painfully with Hyperbole.

It was hard to decide how high-level to make the effect. In the end, I chose to have the effect automatically store things like the state and redirect url in the session, but to ask the developer to store the resulting user session. 

I'm pretty happy with it now. Let me know if you have any thoughts!


adithyaov:
@seanhess Supporting OAuth2 natively is an excellent idea!
I have briefly looked at `Web.Hyperbole.Effect.OAuth2`. Looks good.
I can give more elaborate feedback (if any) after using it.

PR #132, opened 2025-08-12
Add client/dist/live-reload.js to extra-source-files

adithyaov:


seanhess:
Oops, @adithyaov Can you add another PR to make this change to package.yaml? hyperbole.cabal is overwritten automatically by hpack during development.

adithyaov:
Ah, yes, sorry.
Is it possible to add a simple CI to check for `hpack package.yaml`?
It should fail if `hyperbole.cabal` is changed.

adithyaov:
Very likely that this check will be skipped and changes will be lost if there is no automatic check..

seanhess:
I can do a better job of catching it on PRs, but yeah, a CI change would be great

PR #134, opened 2025-08-12
Improve the oauth2 flow

adithyaov:


src/Web/Hyperbole/Effect/OAuth2.hs
```diff
@@ -178,7 +178,7 @@ data TokenType
 instance ToJSON TokenType where
   toJSON s = toJSON $ show s
 instance FromJSON TokenType where
-  parseJSON (String "Bearer") = pure Bearer
+  parseJSON (String ttyp) | T.toLower ttyp == "bearer" = pure Bearer
```
adithyaov, line 181: We need to support the following variations:
- Bearer
- bearer
- BEARER

`token_type` values are case-insensitive, and providers may send any variation. 
For example, GitHub sends `bearer`

seanhess:
Great!

PR #141, opened 2025-08-21
Implement the radio form element and augemnt the example

adithyaov:


src/Web/Hyperbole/HyperView/Forms.hs
```diff
@@ -235,6 +236,16 @@ checkbox isChecked = do
   tag "input" @ att "type" "checkbox" . name nm $ none @ checked isChecked
 
 
+
+radio :: ToParam a => a -> Bool -> View (Input id a) ()
+radio val isChecked = do
+  Input (FieldName nm) <- context
+  let (ParamValue valTxt) = toParam val
```
adithyaov, line 243: I've used `ToParam` here and not `ToJSON` because `gFormParse` uses `FromParam`.

seanhess: Good, forms are url encoded, so that makes sense

examples/Example/Page/FormSimple.hs
```diff
@@ -46,6 +54,23 @@ formView = do
         checkbox False ~ width 32
         label "Favorite?"
 
+    col ~ gap 5 $ do
+      el $ text "Planet"
+      mapM_ (field "planet")
```
seanhess, line 59: This is a misuse of field. It's meant to contain the context of the "name" field of the input. There ought to be a way to use field only once

seanhess (changes requested):

examples/Example/Page/FormSimple.hs
adithyaov: Please see: https://github.com/seanhess/hyperbole/pull/142

PR #142, opened 2025-08-21
Another ambitious experiment with radio

adithyaov:


adithyaov:
@seanhess This seems to be more inline with the idea of how `field` is being used.

adithyaov:
@seanhess, I think the best way is to separate the context features from the label features, like you mentioned above.
This will be a breaking change. 
Changing `label` is okay, although breaking is not silent.
Changing `field` will be a silent breaking change, so we shouldn't do it even in a major release.
We can add a new combinator `name`, instead.

```
name
  :: forall (id :: Type) (a :: Type)
   . FieldName a
  -> View (Input id a) ()
  -> View (FormFields id) ()
name fn inputs =  addContext (Input fn) inputs

label :: View (Input id a) () -> View (Input id a) ()
label = tag "label"
```

adithyaov:
Closing this. Will open another PR after #146 is merged.

PR #149, opened 2025-08-28
Some minor fixes for examples

sidkshatriya:
See commits (very simple fixes)

### DIFF
--- a/examples/Example/View/Layout.hs
+++ b/examples/Example/View/Layout.hs
@@ -65,7 +65,7 @@ sourceLink :: Path -> View c ()
 sourceLink p =
   link sourceUrl "View Source" ~ Style.link
  where
-  sourceUrlBase = [uri|https://github.com/seanhess/hyperbole/blob/0.4/example/|]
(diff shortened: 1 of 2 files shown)


examples/Example/View/Layout.hs
```diff
@@ -65,7 +65,7 @@ sourceLink :: Path -> View c ()
 sourceLink p =
   link sourceUrl "View Source" ~ Style.link
  where
-  sourceUrlBase = [uri|https://github.com/seanhess/hyperbole/blob/0.4/example/|]
+  sourceUrlBase = [uri|https://github.com/seanhess/hyperbole/blob/main/examples/|]
```
sidkshatriya, line 68: In the `0.4` branch which is probably what is live at https://docs.hyperbole.live/simple needs to at minimum change the typo `example` -> `examples` (the links are broken)

sidkshatriya:
@seanhess Gentle ping -- Let me know if these changes don't make sense or you'd like to fix these things in an alternate way...

PR #176, opened 2025-10-24
Add the packcheck CI

adithyaov:
This includes mainly
1. An sdist check - We need to track all the files that are committed
2. Build and test with multiple versions of GHC on different platforms
3. Hlint for lining the .hs files

### DIFF
--- a/.github/workflows/packcheck.yaml
+++ b/.github/workflows/packcheck.yaml
@@ -133,13 +133,13 @@ jobs:
             runner: macos-latest
             cabal_project: cabal.project
 
-          - name: ci
-            command: hlint
-            runner: ubuntu-latest
-            pack_options: >-
-              HLINT_VERSION=3.6.1
-              HLINT_OPTIONS="lint"
-              HLINT_TARGETS="src test examples"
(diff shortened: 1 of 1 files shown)


seanhess (changes requested):

seanhess:
How does packcheck relate to github actions? Will it supersede it? Does github actions run packcheck?

adithyaov:
> How does packcheck relate to github actions? Will it supersede it? Does github actions run packcheck?

We don't use the haskell specific github actions. You can run packcheck locally as well. Packcheck is a script that runs on vanilla linux, mac, and windows. It has a minimal dependencies that most linux/mac distros already come with.

> I'm on the fence about hlint. My editor shows me lints so I never miss them permanently, I'm not sure whether it would be nice if the CI blew up if contributors missed some lints. It's not very important, I wouldn't want a contributor to run away because of it. Do you have a strong opinion?

Sure, we can remove the hlint CI. I don't have a strong opinion on it.

seanhess:
Can you send me some quick instructions on how to use it?

PR #180, opened 2025-11-05
Feedback Requested - HyperView State

seanhess:
I added optional view state to HyperViews.  I'd appreciate some feedback from anyone interested. 

ViewState is entirely backwards-compatible: the default is `()`

```
import Effectful.State.Static.Local

data Counter = Counter
  deriving (Generic)
instance ViewId Counter where
  type ViewState Counter = Int

instance HyperView Counter es where
  data Action Counter
    = Increment
    | Decrement

  update Increment = do
    modify @Int (+ 1)
    pure viewCount
  update Decrement = do
    modify @Int (subtract 1)
    pure viewCount

viewCount :: View Counter ()
viewCount = row $ do
  n <- viewState
  col ~ gap 10 $ do
    el ~ dataFeature $ text $ pack $ show n
    row ~ gap 10 $ do
      button Decrement "Decrement" ~ Style.btn
      button Increment "Increment" ~ Style.btn
```

In the name of simplicity, I've resisted implementing this for a long time. We aren't trying to implement the Elm Achitecture. Instead we have a simpler model that follows the design philosophy of htmx: actions call the server and replace the view with something. However, I don't think it will hurt to have it as an option. 

#### Some outstanding limitations:

* `modify`, `get`, `put`, etc are from `Effectful.State.Static.Local` - It would be useful to export them from somewhere, but they might conflict with other `State` functions, like `Effectful.State.Dynamic`. They also require type annotations. Should we have specialized versions?
* For Technical Reasons (tm) I added it to `ViewId` instead of `HyperView` directly. (It makes working with dumb Views a pain in the ass if they all have to be members of HyperView).


examples/Example/Page/State/View.hs
```diff
@@ -0,0 +1,47 @@
...
+    pure viewCount
+
+viewCount :: View Counter ()
+viewCount = row $ do
+  n <- viewState
+  col ~ gap 10 $ do
+    el ~ dataFeature $ text $ pack $ show n
+    row ~ gap 10 $ do
+      button Decrement "Decrement" ~ Style.btn
+      button Increment "Increment" ~ Style.btn
```
adithyaov: This looks very neat. Are there any limitations you see with this approach?
Alternatively, any consequences?

As I understand it,
The HyperView modifies the state and the view is derived from the state.

It looks like this also changes some idioms. This is no longer htmx way. It's closer to elm like model.

adithyaov (commented):
I've not throughly reviewed the inner wiring but from a user perspective, I like this a lot!
I don't have enough experience with this to see the limitations though.

Apologies for such a delayed feedback.

examples/Example/Page/State/View.hs
adithyaov: I've actually misunderstood the usage,

What I think will be a nice UX:

```haskell
  update Increment prevState = prevState + 1
  update Decrement prevState = prevState - 1
```

In this PR it is currently:

```haskell
  update Increment = do
    modify @Int (+ 1)
    pure viewCount
  update Decrement = do
    modify @Int (subtract 1)
    pure viewCount
```

This is actually a little confusing because of the fact that it supports both types of updates - The htmx way and the elm way.
The bias is towards HTMX-like pattern because of the return type. It's harder to reason about as potentially both the state and the return html can be modified.

Would it be possible keep the older `HyperView` class as is and introduce a new typeclass `ElmView` that only supports elm like pattern?

I really like the pattern of managing state. But it would be best if we are able to introduce it without modifying how the current `HyperView` looks like. Or we modify the current `HyperView` completely.

I'm happy to spend some time on experimenting with this if you think `ElmView` is a good idea.

adithyaov: I'm thinking mostly about the UX. We can also possibly create type aliases for this type of UX.

```
HyperView = HyperViewInternal () View
ElmView s = HyperViewInternal s s
```

I'm thinking out loud - this may not even make sense.

seanhess: I'm 100% sure I don't want to convert `update` into a reducer like elm. I'm also sure I don't want a second typeclass. 

Why would you expect a reducer when Haskell already has a `State` monad/effect? I would guess it's simply because you've worked with Elm and it's intuitive to you, but consider that Elm only does it that way because it _doesn't_ have monads. 

All this PR does is add a build-in state effect to `update` and `view` such that one can easily track view state. The `update` is the same as before: it returns a new view after executing effects. It's just that, like the `Reader viewId` effect that is built-in, there's a second built-in effect `State viewState` that makes it easy to use view state.

adithyaov: > Why would you expect a reducer when Haskell already has a State monad/effect?

My reasoning is more on the lines of constraining the usage.

The `update` has 2 moving parts to it.
1. The return html
2. The state

`HTMX`: Modify html while keeping the state constant
`ELM`: Modify state while keeping the html constant
`??`: Modify both html and state

I want `??` to be undescribable.

I'm happy with using state if we keep the return type of `update` `()` instead of `View ()`.

> Thanks so much for the feedback by the way! It's super helpful to discuss it with you

Glad to be of help!

seanhess:
> @seanhess, Is this change (semantically and functionally to an extent) **exactly** the same as threading state across actions?

Yes exactly. It's a different interface, but it behaves the same way. The framework just handles the threading for you. It would be useful any time you find yourself adding a state parameter to every single action, like with the threaded counter:

```
data Action Counter = Increment Int | Decrement Int
```

adithyaov:
This is a very welcome change :-). I have a much more clear idea of what's going on now.

The only thing I would point out again is that the framework now supports the elm-like idiom.
Although this was possible earlier, it never really occurred to me that I could the framework that way. Or even if it did, I was discouraged by the fact that the author did not intend for actions to be used that way.
With this PR, the framework explicitly supports this use case. 

Although, I don't think this is a bad thing. I would prefer the solution in this PR for state management over threading it in actions.

It's actually great that this is a rather non-invasive change, both in terms of functionality and the usage.
Looking forward to seeing this getting merged!

PR #188, opened 2026-02-04
Fix checkbox input to correctly set checked attribute

fanshi1028:


fanshi1028:
Oh, I should've made the bug more clear. It is different, the checkbox is not checked even "isChecked" is set to "True". Probably due to the style got applied to "none"? This fixed it for me.

PR #200, opened 2026-04-21
js: migrate from `Webpack` to `esbuild`

sectore:
## Why?

Webpack is just used to bundle / minify / watch TS sources. In 2026 we do have better options to replace Webpack for such tasks, which is slow in general. One common option is `esbuild` (simple to use + very fast). Check its [Benchmark details]( https://esbuild.github.io/faq/#benchmark-details) comparing to Webpack.

## Changes

- use `esbuild` to get rid of `Webpack` entirely
- update build/dev scripts
- fix TS warnings: `Boolean` -> `boolean`, `require` -> `import`
- remove all `client/dist/*.d.ts` files (needed if `Hyperbole` client is a JS/TS library, which seems not to be the case).
- update `tsconfig.json`
- update `deps` in `package.json`


seanhess:
Hey @sectore thanks for contributing! I gave it a quick spin - I like the compile speed. 

Two issues need resolving
1. bin/dev compiles once, but doesn't watch the typescript for changes like the old webpack version. SEe below for the 
2. Typescript errors don't show in the console or stop the build process.

Can you figure out how to fix those? 

```
> web-ui@0.6.0 dev
> esbuild src/index.ts --bundle --minify --sourcemap --watch --outfile=dist/hyperbole.js

[watch] stopped automatically because stdin was closed (use "--watch=forever" to keep watching even after stdin is closed)
```

seanhess:
It looks like the webpack `bin/dev` also doesn't recompile on changes. If you can figure out how to get it to work I would be much obliged, but I'll merge this PR as soon as you can resolve the second issue: Typescript errors don't show in the console or stop the build process.

sectore:
Pls try again and let me know.

Another option to get rid of `Webpack` is by using `Vite`, which is using `esbuild` behind the scenes (see https://esbuild.github.io/faq/#production-readiness -> Used by other projects). With that you could use [Vitest](https://vitest.dev/) to test TS sources. I can open another PR for Vite (and maybe another for few tests or so) ...

seanhess:
Webpack isn't causing any problems. Our JS codebase is pretty small, and it comiles quickly enough. I wouldn't want to trade stability for a 100x improvement on compile times when our compile times are less than 0.5 seconds. 

That link makes me feel like esbuild might be too bleeding edge for us. While I think of Hyperbole itself as being ultra-modern, that doesn't mean that it's helpful to do anything other than the most boring and stable thing for the library js. The end-user developer of Hyperbole can still use all the latest hotness in their project, including any custom js. 

What are you thoughts?

sectore:
As long as `Webpack` does a good job for the project, keep it. 

Suggestions (nits): 
- Fix `boolean` typo in TS sources (see [TypeScript -> Basic Types]( https://www.typescriptlang.org/docs/handbook/basic-types.html))
- keep dependencies in `package.json` up-to-date
- Remove `Webpack` from `flake.nix` OR `package.json`. It will be installed twice by running `npm i` (as described in [local development](https://github.com/seanhess/hyperbole/blob/main/docs/dev.md)) in `nix-store` AND `node_modules`) with two different versions.
- _I was wrong with `Vite` + `esbuild`._ Vite is not using `esbuild` anymore, it _"uses [Rolldown](https://rolldown.rs/) and [Oxc](https://oxc.rs/) based tools instead of `esbuild` and `Rollup`."_ ([source]( https://github.com/vitejs/vite/blob/640202a2167b0c19b94e4d3b8ff87309ae1f44d0/docs/guide/migration.md#rolldown))
- _Side note:_ Even if JS sources in Hyperbole are small, they are very important to provide the bridge to get updates in the browser etc. It might be worth to add linter, formatter and tests for JS sources, too (all things you have already done with Haskell sources). JS innovations are there to make things better (not worse), many of them have already been proven very well (e.g. `Vite` & Co). `Webpack` is a stale thing. 

Feel free to close current PR. 

Love the project btw.

seanhess:
@sectore super helpful suggestions and context, thank you. I agree, the JS is important and ought to be tested thoroughly. I'll lift a couple changes from this PR manually, but I would absolutely merge a PR with client-side tests and some more discipline. 

And thanks! I'm glad you're enjoying it!

seanhess:
I'm going to go ahead and merge this. Now that it is reporting errors it works great. It wouldn't be hard to go back to webpack if we have problems

PR #201, opened 2026-04-21
demo: add gh + hackage icons

sectore:
Currently important links to GitHub / Hackage are located at the bottom of the intro page only.  

How about to add these to the menu using icons, too? So lazy users (like me) will find it more easily...

preview-icons


seanhess:
Love it, thanks!

PR #202, opened 2026-04-29
toActionInput throws input holes, fixes #194

### LINKED ISSUE #194: toActionInput evaluates undefined
Consider the following example:

```haskell
instance HyperView TestView es where
    data Action TestView = Test Int Int

view :: View TestView ()
view = dropdown (\x -> Test x 1) 0 none
```

In this case, `toActionInput` will evaluate undefined in `toAction (act undefined)`, which only seems to work if the supplied argument is the last one for the given action, i.e., `\x -> Test 1 x` works but `\x -> Test x 1` does not.

seanhess:
See #194 

Instead of using undefined to serialize a partially applied constructor, we `throw InputHole` - we catch it during serialization and turn it into a `Hole`. 

This inspired a refactor to remove Web.Hyperbole.Data.Param. Everything now uses ToJSON as a generic serialization format, then encodes in its own way. This changes QueryData, Session, and Encoded (ViewId, ViewAction)


seanhess:
What about forcing the user to be more explicit? It might help users avoid doing crazy things with `(input -> Action view)`

In the latest push, `onInput` `(Text -> Action view)` works out of the box. Other types must implement `UserInput`. It doesn't assume it knows how you want to parse numbers from an `<input>`, for example. 

```
-- From Web.Hyperbole.Data.Argument
class UserInput a where
  parseInput :: Text -> Either String a
  default parseInput :: (FromJSON a) => Text -> Either String a
  parseInput = decodeArgument


instance UserInput Text where
  parseInput = pure


instance {-# OVERLAPPABLE #-} (UserInput a) => UserInput (Maybe a) where
  parseInput "" = pure Nothing
  parseInput t = Just <$> parseInput @a t


instance {-# OVERLAPS #-} UserInput (Maybe Text) where
  parseInput = pure . Just

-- Examples: 


-- If you want another type as an input, you must implement UserInput. We don't automatically parse integers from search fields
newtype NumInput = NumInput Int
  deriving (Generic, Show, Eq)
  deriving anyclass (UserInput, FromJSON, ToJSON)

data Custom = Custom 
  deriving (Generic, ToJSON, FromJSON, UserInput)

data NotAnInput = NotAnInput Int Text
  deriving (Generic, ToJSON, FromJSON)

data SomeAction
   = IntegerInput Int NumInput
   | OutOfOrder NumInput Int
   | TextInput Text
   | NullableInput (Mabye Text)
   | CustomInput Custom
   | BadInput Int
   | BadInput2 NotAnInput
 
 -- These work
toActionInput TextInput
toActionInput NullableInput
toActionInput (IntegerInput 4)
toActionInput (CustomInput)
toActionInput (\n -> OutOfOrder n 4)
 
 -- Throws a runtime error
 toActionInput (NullableInput . Just)
 toActionInput (Nested . NumberPair 0)
 
 -- Compile Errors
toActionInput BadInput
toActionInput (NotAnInput 3)
toActionInput (\n -> NotAnInput n "hello")
toActionInput (BadInput . (*2))
toActionInput (IntegerInput 4 . (*2))
```

What do you think?  How could it be improved? I think this might make people think twice before doing crazy things with the partial constructor.

seanhess:
@kschweppe do you have any thoughts on this solution? I could use a second set of eyes on this.

kschweppe:
Sure! Sorry for the delay, I just didn't get to have a closer look over the weekend.

I generally agree that it should be more explicit, and was also thinking we need a typeclass for this.
I still need to play around a bit more, but this seems to work well and cover most cases.

Is there any drawback in providing a default implementation for numbers? I guess having number dropdowns is a common use case, for example I have dropdowns for year and month and needed to add NumInput to both.

And then I'm wondering if this can/should also work with search input, which actually was my initial use case.
So I wanted something like `search (SearchUser . SearchInput False)` (which now at least throws a runtime error). Do you think it makes sense to use `UserInput` for other events as well, like `onInput`, or only for `onChange`?

seanhess:
@kschweppe Ok, more thoughts:

The developer needs control of how to handle errors when parsing raw user text input. That means that converting raw text (search, onInput, onChange) shouldn't be automatically converted to a user type by the framework like we were thinking. There should be at least 1 line of code in the update handler to parse it into anything but text. 

That leaves us with two options:

1. `onChange` uses Text only just like `onInput`. So constructors must be `SomeAction Text` not `SomeAction AnyType`. To handle integers, you could attempt to read it with `Text.Read.readMay` and display the error however you like, for example, with `Web.Hyperbole.Effect.Response.parseError`. 

But let's think about how we currently handle other unstructured or partially structured info in the action: forms, queries, sessions, etc. We don't put that data on the actions: it's on the Request, and you use a function like `session`, `query`, `param`, or `formData` to convert it into your type. That brings us to another idea: 

2. add a `userInput` function that parses the type for you. This might use a typeclass `UserInput` with a parsing function `(Text -> Either String mytype)`. Then the input isn't embedded into the action at all. 

This completely sidesteps this entire issue and PR. If user input isn't appended to a partial constructor, we don't have to worry about trying to serialize one and hydrating it later. User Input would look just like button input, but the developer would use an extra function to grab the input. 

Here's an example of how things work today:

```
data CustomText = CustomText
  deriving (Generic, ViewId)

instance HyperView CustomText es where
  data Action CustomText
    = ShowInput Text

  update (ShowInput t) =
    pure $ viewCustom t

customText :: View CustomText ()
customText = do
  search ShowInput 250 ~ border 1 . pad 10 @ placeholder "Type something"

viewCustom :: Text -> View CustomText ()
viewCustom t = row $ do
  col ~ gap 10 $ do
    customText
    case t of
      "" -> none
      _ -> el $ text $ "You typed: " <> t
```

If we adopted option 2, it would change to look like this:

```
instance HyperView CustomText es where
  data Action CustomText
    = ShowInput

  update ShowInput =
    t <- userInput
    pure $ viewCustom t
```

So the big question is: Does user input belong somewhere in the action constructor, or is it better to have the developer manually bring it in via a function like `formData`. Any thoughts?

seanhess:
NEED FEEDBACK! (I'll summarize the decision now)

Currently, interaction with a `change` or `input` event puts `input.value` on the action, like this: 

```
instance HyperView SearchTest es where
  data Action CustomText
    = Search Text

  update (SearchTest term) =
    pure $ viewSearchResults term
```

I'm proposing changing this to include `input.value` in the body of the message instead, which will simplify the code, prevent some runtime errors, and also make the interface match how forms work (via `formData`). 

```
instance HyperView SearchTest es where
  data Action CustomText
    = Search

  update SearchTest =
    term <- userInput -- or inputValue?
    pure $ viewSearchResults term
```

kschweppe:
All right! I think out of these two options, I definitely prefer the second.
It simplifies things a lot, there are no unexpected runtime errors, and if only text arguments are supported it also isn't more verbose to use.

The original approach is only more ergonomic with automatic conversion.
I would say for a static dropdown it's reasonable to assume that everything is controlled by the framework and no parse errors are to be expected. But yes, otherwise parse errors should be handled, and maybe this exception doesn't justify the additional complexity.

sectore:
+1 for simplification (TS: `encodedParam` -> `encodedAction`, HS: `ToEncoded,FromEncoded,ToParam,FromParam` -> `ToJSON,FromJSON` etc.), for preventing possible runtime errors and for updating tests (specs).

seanhess:
Closing in favor of #211

PR #204, opened 2026-04-30
Vite: build, test, format, lint

sectore:
With this PR, Hyperbole gets everything needed for solid JS/TS development. 

_Note:_ Check commits step by step to see what is needed. _Code changes look much more as it is._

## Changes

- [`Vite 8`](https://vite.dev/config/build-options#build-lib) ([release notes](https://vite.dev/blog/announcing-vite8)) to bundle + minify + watch (dev only) files.
- [`Vitest`](https://vitest.dev/guide/features.html#watch-mode) to run tests
- [`oxlint`](https://oxc.rs/docs/guide/usage/linter.html) for linting. From [same builders](https://voidzero.dev/) as Vite.
- [`oxfmt`](https://oxc.rs/docs/guide/usage/formatter.html) for formatting. From [same builders](https://voidzero.dev/) as Vite.
- Get rid of `esbuild` as Vite handles bundling and the rest now.
- Use latest `Node 24`. It's needed to run `Vite 8` etc. A separate `nixpkgs-node` input allows updating Node to 24 via `nix flake update nixpkgs-node` without updating/changing/breaking the Haskell build. See e212f356179b3df1474c01af9f2ee8d885c2be6b for details.


client/src/action.ts
```diff
@@ -1,92 +1,87 @@
...
-  state?: ViewState
-  meta: Meta[]
-  form: URLSearchParams | undefined
-}
-
-
-
-
-export function actionMessage(id: ViewId, action: EncodedAction, state: ViewState | undefined, reqId: RequestId, form?: FormData): ActionMessage {
+  viewId: ViewId;
```
seanhess: Does the linter require semicolons? Can we turn that off and remove them? I view them as useless extra noise.

client/src/action.ts
```diff
@@ -1,92 +1,87 @@
...
 }
 
 export function renderActionMessage(msg: ActionMessage): string {
-  let header = [
-    "|ACTION|",
-    "ViewId: " + msg.viewId,
-    "Action: " + msg.action,
-  ]
-
+  let header = ["|ACTION|", "ViewId: " + msg.viewId, "Action: " + msg.action];
```
seanhess: I prefer the multi-line version

client/src/action.ts
```diff
@@ -1,92 +1,87 @@
...
   }
 
-  header.push("RequestId: " + msg.requestId)
+  header.push("RequestId: " + msg.requestId);
 
-  return [
-    header.join('\n'),
-    message.renderMetas(msg.meta),
-  ].join('\n') + renderForm(msg.form)
+  return [header.join("\n"), message.renderMetas(msg.meta)].join("\n") + renderForm(msg.form);
```
seanhess: Same as above. It's useful to visually separate the code with newlines to reflect the message being separated by them

seanhess (changes requested):

seanhess:
Cool! Thanks for doing this! 

1. See the review. Could you remove the extra semicolons? 
2. Will you look into the build failure? I'm not sure why it wants those lint files in extra-source-files, when it doesn't seem to care about the rest of the typescript source.

client/src/action.ts
sectore: You can add any format rules as you like: https://oxc.rs/docs/guide/usage/formatter/config.html#configuration-file-format

Complete list of options: https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html

Do you wanna try?

client/src/action.ts
sectore: https://github.com/seanhess/hyperbole/pull/204#discussion_r3170264557

client/src/action.ts
sectore: https://github.com/seanhess/hyperbole/pull/204#discussion_r3170264557

client/src/action.ts
sectore: I could not find any setting to keep a multi-line format for arrays. Even `prettier` (another well-know formatter) seems not to support it by default. Prettier does need an [extra (third-party) plugin](https://github.com/electrovir/prettier-plugin-multiline-arrays/tree/dev) for doing that. 

Optional: To keep the format add `prettier-ignore` above the line (see oxfmt ->[prettier-compatibility](https://oxc.rs/docs/guide/usage/formatter/ignore-comments.html#prettier-compatibility)). 

_Just wondering:_ Does `Fourmolu` support something like that for Haskell code (_I could not find such option there ..._)?

client/src/action.ts
sectore: It seems there is no way to format `multi-line arrays` in such a way. See https://github.com/seanhess/hyperbole/pull/204#discussion_r3172792472

sectore:
>     1. See the review. Could you remove the extra semicolons?
- Semicolons: Done. 
- Other format change requests: Pls check my comments. 
 
>     2. Will you look into the build failure? I'm not sure why it wants those lint files in extra-source-files, when it doesn't seem to care about the rest of the typescript source.
- `extra-source-files` issue: Fixed. 
- Another error at CI: Based on introducing `selda-sqlite`? It's currently part of latest `main` only, same issue at CI there.

seanhess:
Looks good, thanks! I'll play with the formatter config a bit.

PR #222, opened 2026-05-21
ci(haskell): lint + format

sectore:
Note: Most changes (99%) in `*.hs` files are just formatting changes.

Question: Why are two `fourmolu.yaml` files defined in root and in `demo` folders? The only difference is `newlines-between-decls`  `2` vs. `1`. 

Having two config files will let `fourmolu` break on CI. It seems it will pick up one config only there. That's why I have removed `fourmolu.yaml` from `demo`. That's the main reason to have many formatting changes in this PR.


seanhess:
> Question: Why are two fourmolu.yaml files defined in root and in demo folders? The only difference is newlines-between-decls 2 vs. 1.

Because 2 lines of whitespace is too much for documentation embedded by docgen, but is easier on the eyes for normal development.

sectore:
> Because 2 lines of whitespace is too much for documentation embedded by docgen, but is easier on the eyes for normal development.

What about to use `newlines-between-decl: 1` for all files to have one `fourmolu.yaml` only?

seanhess:
> What about to use newlines-between-decl: 1 for all files to have one fourmolu.yaml only?

I don't know... 1 newline feels really dense to me when working in the main library. 

Why does it break in CI?

sectore:
> Why does it break in CI?

It seems `hlint` uses one config file to run against all files at CI. Maybe there is a way to setup different `hlint` jobs for different folders and config files. However, users might still run into same issue by running `hlint` locally without pointing to different configs.

seanhess:
Let's figure out a workaround. Nested fourmolu files are definitely supported: https://fourmolu.github.io/config/ - that's exactly how you override a higher-up config with something specific to a subdirectory

> Fourmolu looks for a fourmolu.yaml file in **all parents** of the current directory

sectore:
> Let's figure out a workaround. 

Ok, two `haskell-actions/run-fourmolu` actions run now. CI is happy. Just few formatting / linting changes in the PR now.

PR #225, opened 2026-05-22
update/fix `filter` example 

sectore:
- [x] clear input on `ESC` 
- [x] add `back` button
- [x] ~~Selecting `Any` from "Language Family"  breaks round-trip.~~

   ~~Error: "Parse Input: parsing Example.Data.ProgrammingLanguage.LanguageFamily failed, expected String, but encountered Null".~~
   
   ~~Try it [here](https://hyperbole.live/data/filter).~~

   hb-filter-any
   
   ~~**Fix:** Now [`<option>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/option) will be rendered to `<option value="">` in case it gets a `Nothing` or any other `Null` like value.~~

### DIFF
--- a/src/Web/Hyperbole/HyperView/Input.hs
+++ b/src/Web/Hyperbole/HyperView/Input.hs
@@ -3,7 +3,7 @@
 
 module Web.Hyperbole.HyperView.Input where
 
-import Data.Aeson (FromJSON, ToJSON (toJSON), Value (Null))
+import Data.Aeson (FromJSON, ToJSON)
 import Data.String.Conversions (cs)
 import Data.Text (Text)
 import GHC.Generics (Generic)
@@ -55,9 +55,7 @@ option
   -> View (Option id opt) ()
 option opt cnt = do
   os :: Option id opt <- viewId
-  -- For "null" like arguments such as `Nothing` set `value=""` (but never `value="null"`, which breaks the round-trip)
-  let val = if toJSON opt == Null then "" else encodeArgument opt
-  tag "option" @ att "value" val @ selected (os.defaultOption == opt) $ text cnt
+  tag "option" @ att "value" (encodeArgument opt) @ selected (os.defaultOption == opt) $ text cnt
 
 
 -- | sets selected = true if the 'dropdown' predicate returns True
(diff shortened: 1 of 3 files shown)


src/Web/Hyperbole/HyperView/Input.hs
```diff
@@ -55,7 +55,9 @@ option
   -> View (Option id opt) ()
 option opt cnt = do
   os :: Option id opt <- viewId
-  tag "option" @ att "value" (encodeArgument opt) @ selected (os.defaultOption == opt) $ text cnt
+  -- For "null" like arguments such as `Nothing` set `value=""` (but never `value="null"`, which breaks the round-trip)
+  let val = if toJSON opt == Null then "" else encodeArgument opt
```
seanhess: This was a bug with `InputValue`, fixed in https://github.com/seanhess/hyperbole/commit/483d5016dc7ad540e8355a4e95d5802a2609258a

Can you remove this change?

seanhess (changes requested):

src/Web/Hyperbole/HyperView/Input.hs
sectore: Reverted.

Works now. However, I would expect to have `value=''` (default for **non** values) instead of value='null' (which is still a valid value). Imagine a list of options, where one value should be "null" (string)... 

Check in `Filter` example:

Image

seanhess: That's true... We use `encodeArgument`. When used as an actual `Encoded` argument, `null` makes lots of sense (and an empty string would break decoding)

An empty string does make more sense as an empty value for options though. What do *forms* need? Options have to work for both. I bet they want the empty string too...

I'll double check and maybe add some tests for the Form case. But maybe we want this change after all

seanhess: Sure enough, testing the round trip between option and parsing a form field failed with "null". I stole your fix and committed it with the tests in https://github.com/seanhess/hyperbole/commit/d88ca89ddbe318f7f15c706025e78471d549cc12

